PR #2, opened 2017-01-30
Add Felix's build dependencies.

fasaxc:
Also, direct entrypoint's own output to stderr to make it easier to port targets that use pipes.


fasaxc:
@neiljerram desire is to standardise on one build container for golang; but that means it needs to have all our deps in it.  This one was based on the one from calicoctl + the one from felix so it's almost there.

PR #4, opened 2017-02-24
Add Go standard library patch to use VFORK for exec.Command().

fasaxc:
For more details: https://github.com/golang/go/issues/5838.


tomdee:
Could you add a link to the Go issue in the Dockerfile? 
I have a slight concern that this will hamper our ability to upgrade our Go version, but I think the cost is worth it. So happy to merge once you add that link in the Dockerfile.

fasaxc:
Done, also, it looks like the patch will land in 1.9, so if the patch starts failing, that probably means it's no longer needed :-)

PR #6, opened 2017-04-10
Rev go to 1.8.1 and pin gometalinter to working version.

fasaxc:
Switch to using go 1.8.1 base image.  Had to pin gometalinter because master currently seems to be broken.


tomdee:
LGTM

PR #9, opened 2017-05-25
enable support for ppc64le arch

ligc:
Enable support for ppc64le arch, changes:
* new Dockerfile.ppc64le
* new build target calico/go-build-ppc64le in Makefile


Dockerfile.ppc64le
```diff
@@ -0,0 +1,39 @@
...
+
+# Install tool to merge coverage reports.
+RUN go get github.com/wadey/gocovmerge
+
+# Install tool for continuous code coverage tracking
+RUN go get github.com/mattn/goveralls
+
+# Ensure that everything under the GOPATH is writable by everyone
+RUN chmod -R 777 $GOPATH
+
```
fasaxc: Is there a reason you're not including the entrypoint script here?  We use that to make sure that file ownership is sensible (i.e. to stop all the build artefacts being owned by root on the host system).

Makefile
```diff
@@ -1,2 +1,5 @@
 calico/go-build:
 	docker build -t calico/go-build .
+
+calico/go-build-ppc64le:
+	docker build -t calico/go-build-ppc64le -f Dockerfile.ppc64le .
```
fasaxc: Please can you comment the requirements for running this?  Presumably it needs to be run on a PPC-based system?

Dockerfile.ppc64le
```diff
@@ -0,0 +1,39 @@
+FROM ppc64le/golang:1.8.1
```
fasaxc: Is there a golang-alpine for PPC?  Feels like it'd be nice to be consistent if we can.  We try to use alpine because it's very light weight.

ligc: nope, we do not have golang-alpine for PPC. the ppc64le/golang:1.8.1 is based on Debian base. Actually the alpine has not been ported to PPC yet.

Makefile
ligc: yes, needs to run on a PPC-based system or in a boot2docker for PPC environment.

Dockerfile.ppc64le
ligc: the su-exec does not exist in the ppc64le/golang:1.8.1 image, let me look further into this. Thanks for the comments.

ligc:
Dockerfile does not support INCLUDE for now, if we are concerned with the duplicate information in Dockerfile and Dockerfile.ppc64le, we could create some scripts to generate the Dockerfile and Dockerfile.ppc64le from some template files.

BTW, I started a topic in the docker community about how to manage the duplicate information in different Dockerfiles, https://forums.docker.com/t/dockerfile-for-different-archs/32711. FYI.

ligc:
@fasaxc For generating the Dockerfile and Dockerfile.ppc64le, how about we use something like this in Makefile:

```
dockerfile: 
    export ARCH=x86_64
    export BASE-IMAGE=golang:1.8.1
    ./Dockerfile.template > Dockerfile

dockerfile-ppc64le
    export ARCH=ppc64le
    export BASE-IMAGE=ppc64le/golang:1.8.1
    ./Dockerfile.template > Dockerfile.ppc64le
```

The Dockerfile.template is a template and bash script that honors the env vars and generate the Dockerfile/Dockerfile.ppc64le.

Does this make sense?

Dockerfile.ppc64le
ligc: latest updates, IBM just made the ppc64le/alpine available at https://hub.docker.com/r/ppc64le/alpine/,  I am working with the team to create the ppc64le/golang image, when it is ready, hopefully the only difference between Dockerfile and Dockerfile.ppc64le will be the FROM statement for base image.

ligc:
IBM made the ppc64le/alpine:3.6 and ppc64le/golang:1.8.1-alpine available on github. Here the is the new logic:

1. Added the shell script create_dockerfile.ppc64le.sh to create the Dockerfile.ppc64le on the fly based on the Dockerfile, since the Dockerfile and Dockerfile.ppc64le both uses the alpine golang image, so we could keep almost all the Dockerfile content in Dockerfile.ppc64le, the only two changes are the base image name and the syscall patch(which is for x86 only, not for ppc64le).

2.  Update Makefile to add "calico/go-build-ppc64le" to build the calico/go-build-ppc64le

@fasaxc Could you help review the changes again? thank you.

create_dockerfile.ppc64le.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# Create Dockerfile.ppc64le based on Dockerfile
```
fasaxc: Please use `set -e` so this script will exit on failure, then I recommend making the changes to a temp file and moving it into place once it is ready.  That ensures that the file will only be moved into place on success so make will correctly rebuild it after a failure (for example, caused by Ctrl-C).

ligc: @fasaxc  Thanks for the comments, I updated create_dockerfile.ppc64le.sh  based on your comments.

djlwilder:
Hi everyone,  sorry for joining this discussion so late.
In my opinion ligc's approach of a separate Dockerfiles per arch is the way to go.   Dockerfile creates a base for the arch therefore any change must be considered separately for each arch.  Having separate files will forces that review. The "sed the file" approach is great when only one line changes but as time goes on it will become hard to maintain. Per-Arch Dockerfiles have been adopted by several other projects including Docker.

fasaxc:
@ligc WDYT to @djlwilder's idea?

ligc:
@fasaxc David has rich experience with enabling ppc64le support for different open source projects, he mentioned in an email to me that "Most other project (Docker for one) is using a separate Dockerfile for each supported arch", I agree with what David suggested, i.e., to use a separate Dockerfile.ppc64le.  If everyone agree, I will update the code to use a separate Dockerfile.ppc64le. Thanks.

fasaxc:
Sounds good to me.

ligc:
Thanks @fasaxc , @djlwilder for the review and comments, I updated the code based on the discussion. Please help review the new changes.

PR #11, opened 2017-07-25
Switch ppc64le build environment to ppc64le/golang

djlwilder:
The Go compiler on ppc64le/golang:1.8.1-alpine is built to generate PIEs
by default (buildmode=pie). Unfortunately this is inconsistent with x86
Alpine that defaulting to buildmode=exe.  Therefor I am switching the
ppc64le build environment to ppc64le/golang:1.8.3. This environment has
proven to be stable on an internal ppc64le Calico port.


fasaxc:
Sorry @djlwilder, I didn't spot this until @caseydavenport pointed me at it.  Does PIE actually cause any problems?  It'd be good to understand what we're doing wrong if so!

djlwilder:
Hi @fasaxc 
I hit a couple of issues trying to build calico as PIE. They are fixable but  for now I am attempting to be consistent between archs.   If we decide to support PIE builds in the future then we need to make a few changes to a number of the Makefile.

For example, in felix/Makefile the following checks are made.
( ldd bin/calico-felix 2>&1 | grep -q -e "Not a valid dynamic program" \
-e "not a dynamic executable" || \
 ( echo "Error: bin/calico-felix was not statically linked"; false ) )'

ldd will report a PIE executable as dynamic linked and the build will fail.  Several of the Makefiles have the same check.  We can just remove all these checks.

We can build static in the Alpine ppc environment by adding a bunch of flags as so.
 # go build -buildmode exe -ldflags "-linkmode external -extldflags -static" hi.go
# file hi
hi: ELF 64-bit LSB executable, 64-bit PowerPC or cisco 7500, version 1 (SYSV), statically linked, not stripped, with debug_info

I tried this out with the felix repo and hit a number of other issues with various go utilities ( sorry i don't recall the details).

djlwilder:
@fasaxc: The latest ppc64le Alpine images have resolved this issue, go build now produces static binaries equivalent to the amd64 image.  Closing out this PR#.

PR #12, opened 2017-08-15
Added docker for recipes that may build inside a container

mgleung:
Some builds require their dependencies to be built from a separate image.  To allow for this to happen within go-build we need docker for those cases.


heschlie:
LGTM

PR #15, opened 2017-10-31
Rev to go 1.8.5 and make sure we pull in security updates.

fasaxc:


Dockerfile.ppc64le
```diff
@@ -1,4 +1,4 @@
-FROM ppc64le/golang:1.8.1-alpine
+FROM ppc64le/golang:1.8.5-alpine3.6
```
neiljerram: add `apk update --no-cache` to this one too?

PR #20, opened 2017-11-22
Add support for Arm64 server

TrevorTaoARM:
Makefile is updated to add a build command for arm64
Add a Dockerfile.arm64 for arm64 docker image building.

Signed-off-by: Trevor Tao <trevor.tao@arm.com>

### DIFF
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,4 +1,4 @@
-FROM arm64v8/golang:1.8.1-alpine
+FROM arm64v8/golang:1.9.2-alpine3.6
 MAINTAINER Trevor Tao <trevor.tao@arm.com>
 
 # Install su-exec for use in the entrypoint.sh (so processes run as the right user)
@@ -10,15 +10,8 @@ MAINTAINER Trevor Tao <trevor.tao@arm.com>
 # Install wget for fetching glibc
 # Install make for building things
 # Install util-linux for column command (used for output formatting).
-RUN apk add --no-cache su-exec \
-  bash \
-  curl \
-  git \
-  make \
-  mercurial \
-  openssh \
-  util-linux \
-  wget
+RUN apk add --no-cache su-exec curl bash git openssh mercurial make wget util-linux docker tini
+RUN apk upgrade --no-cache
 
 # Disable ssh host key checking
 RUN echo 'Host *' >> /etc/ssh/ssh_config \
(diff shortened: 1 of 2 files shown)


Dockerfile.arm64
```diff
@@ -0,0 +1,79 @@
+FROM arm64v8/golang:1.8.1-alpine
```
fasaxc: We recently moved to 1.9, and it looks like there's a `arm64v8/golang:1.9` now, could you switch to that?

Dockerfile.arm64
```diff
@@ -0,0 +1,79 @@
...
+# Install su-exec for use in the entrypoint.sh (so processes run as the right user)
+# Install bash for the entry script (and because it's generally useful)
+# Install curl to download glide
+# Install git for fetching Go dependencies
+# Install ssh for fetching Go dependencies
+# Install mercurial for fetching go dependencies
+# Install wget for fetching glibc
+# Install make for building things
+# Install util-linux for column command (used for output formatting).
+RUN apk add --no-cache su-exec \
```
fasaxc: Looks like you've lost some of the installs here (and the upgrade at the end).  I know some of the builds rely on the docker binary, for example.  Please can you add those back?

Dockerfile.arm64
```diff
@@ -0,0 +1,79 @@
...
+
+# Install glibc
+RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub \
+  && wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.23-r3/glibc-2.23-r3.apk \
+  && apk add glibc-2.23-r3.apk
+
+# Disable cgo so that binaries we build will be fully static
+ENV CGO_ENABLED=0
+
+# Apply patches to Go runtime and recompile
```
fasaxc: Please remove this comment, you're not applying the patch.  (With go v1.9 the patch has been upstreamed anyway.)

Dockerfile.arm64
```diff
@@ -0,0 +1,79 @@
...
+
+# Install license checking tool
+RUN go get github.com/pmezard/licenses
+
+# Install tool to merge coverage reports
+RUN go get github.com/wadey/gocovmerge
+
+
+# Install patched version of goveralls (upstream is bugged if not used from Travis).
+RUN go get -u -d github.com/fasaxc/goveralls && \
```
fasaxc: Why does this need to be different in the ARM version?

Dockerfile.arm64
```diff
@@ -0,0 +1,79 @@
...
+  && ln -s $(which gometalinter.v1) /usr/local/bin/gometalinter \
+  && gometalinter --install
+
+# Install license checking tool
+RUN go get github.com/pmezard/licenses
+
+# Install tool to merge coverage reports
+RUN go get github.com/wadey/gocovmerge
+
+
```
fasaxc: I think you're missing the yaml tool, which some of our builds rely on.

TrevorTaoARM:
@fasaxc 
I have updated the Dockerfile.arm64 to all your comments, and I think the missing parts could be found now. Would you please review it again?

For the 'goveralls', the origin binary in the 'Dockerfile' is for x86, so it has to be built from source for arm64.

Thank you very much

Issue #24, opened 2018-02-06
Cross-compile calico for arm64 and ppc64le from amd64


deitch:
Following the merge of #23 , which simplifies building go-build for arm64 on arm64 and ppc64le on ppc64le, this issue is intended as a tracking issue for cross-compile binaries and cross-build of docker images for arm64 and ppc64le (in addition to amd64, of course) from amd64.

Since every CI-as-a-service provides amd64 only, we want to be able to run the entire build and push from amd64. 

Related: see https://github.com/projectcalico/go-build/issues/22

cc @caseydavenport @tomdee @fasaxc 

Ideas and issues raised by the above people are in a follow-on comment.

deitch:
There are two ways to compile from amd64 to arm64:

* `GOARCH=aarch64 go build ...`
* run in a container with `qemu` providing an emulated architecture. See https://blog.hypriot.com/post/setup-simple-ci-pipeline-for-arm-images/ as posted by @liljenstolpe and @ahrkrak 

The first option generally is preferred, but will not always work. Notably, anything with `CGO` may be an issue.
Additionally, if we need to install additional packages using `apk`, we may have issues.

Note that `apk` does support copying files into alternate locations for alternate arch, using `--arch <arch> --root <root>`, e.g. `apk --arch aarch64 --root /var/alt/pkg/aarch64/ add <package>`. This may provide a viable option, especially with multi-stage container builds.

An example of emu-based building is from flannel, notably [here](https://github.com/coreos/flannel/blob/ff1b89d755d5fe41add70ad6e5e75f1a9aae9de5/Makefile#L48)

djlwilder:
I don't see the need  to install packages for the non-native arch into the container.  Everything executed should be amd64 amd64 with the exception of binaries we just built and sometimes we need to do that  therefore qemu is needed in the go-build:amd64 container.  

I played with this approach and will push a RFC to go-build to add qemu-statuc-user.  Using this update I was able to build a ppc64le felix container on amd64.  It takes a bunch of other changes to felix to as well make this work  but its a start.

deitch:
Let's move the discussion over there?

PR #25, opened 2018-02-07
Add qemu-user-static into the amd64 go-build container.

djlwilder:
You can cross-build golang applications by setting GOARCH=<somearch>,
However, we often need to execute the resulted binary as part of the
build process, here we use qmeu-user-static as the interpreter.  Both
ppc64le and aarm64 qmeu interpreters are installed into the amd64 image.
I also create a writable directory for golang to install the arch
run-times when needed.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -10,7 +10,7 @@ MAINTAINER Tom Denham <tom@projectcalico.org>
 # Install wget for fetching glibc
 # Install make for building things
 # Install util-linux for column command (used for output formatting).
-RUN apk add --no-cache su-exec curl bash git openssh mercurial make wget util-linux docker tini
+RUN apk add --no-cache su-exec curl bash git openssh mercurial make wget util-linux docker tini file
 RUN apk upgrade --no-cache
 
 # Disable ssh host key checking
@@ -58,6 +58,14 @@ RUN go get github.com/mikefarah/yaml
 RUN wget https://github.com/fasaxc/goveralls/releases/download/v0.0.1-smc/goveralls && \
     chmod +x goveralls && \
     mv goveralls /usr/bin/
+
+# Install qemu user static for all our supported architectures.
+COPY bin/qemu-*-static /usr/bin/
+
+# When running cross built binaries run-times will be auto-installed,
+# insure the install directory is writable by everyone.
+RUN mkdir -m +w -p /usr/local/go/pkg/linux_ppc64le
+RUN mkdir -m +w -p /usr/local/go/pkg/linux_arm64
 
 # Ensure that everything under the GOPATH is writable by everyone
 RUN chmod -R 777 $GOPATH
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -10,6 +10,11 @@ MAINTAINER Trevor Tao <trevor.tao@arm.com>
 # Install wget for fetching glibc
 # Install make for building things
 # Install util-linux for column command (used for output formatting).
+
+# Enable non-native builds of this image on an amd64 hosts.
+# This must be positioned as the first executable command in this file! 
+COPY  bin/qemu-aarch64-static /usr/bin/qemu-aarch64-static
+
 RUN apk add --no-cache su-exec curl bash git openssh mercurial make wget util-linux docker tini
 RUN apk upgrade --no-cache
 
--- a/Dockerfile.ppc64le
+++ b/Dockerfile.ppc64le
@@ -10,6 +10,11 @@ MAINTAINER David Wilder <wilder@ibm.com>
 # Install wget for fetching glibc
 # Install make for building things
 # Install util-linux for column command (used for output formatting).
+
+# Enable non-native builds of this image on an amd64 hosts.
+# This must be positioned as the first executable command in this file! 
+COPY  bin/qemu-ppc64le-static /usr/bin/qemu-ppc64le-static
+
 RUN apk add --no-cache su-exec curl bash git openssh mercurial make wget util-linux tini
 RUN apk upgrade --no-cache
 
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,11 @@ else
 maybedefault:
 endif
 
+.PHONY: register
+# Enable binfmt adding support for miscellaneous binary formats.  This is needed for building non-native images.
+register:
+	sudo docker run --rm --privileged multiarch/qemu-user-static:register || true
+
 bin/qemu-ppc64le-static:
 	mkdir bin || true
 	wget https://github.com/multiarch/qemu-user-static/releases/download/v2.9.1/qemu-ppc64le-static.tar.gz 
@@ -35,7 +40,7 @@ bin/qemu-aarch64-static:
 
 build: calico/go-build 
 
-calico/go-build: bin/qemu-ppc64le-static bin/qemu-aarch64-static
+calico/go-build: bin/qemu-ppc64le-static bin/qemu-aarch64-static register
 	# Make sure we re-pull the base image to pick up security fixes.
 	docker build --pull -t $(ARCHIMAGE) -f $(DOCKERFILE) .
 
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,21 @@ else
 maybedefault:
 endif
 
-build: calico/go-build
+bin/qemu-ppc64le-static:
+	mkdir bin || true
+	wget https://github.com/multiarch/qemu-user-static/releases/download/v2.9.1/qemu-ppc64le-static.tar.gz 
+	tar zxvf qemu-ppc64le-static.tar.gz -C bin
+	rm qemu-ppc64le-static.tar.gz
 
-calico/go-build:
+bin/qemu-aarch64-static:
+	mkdir bin || true
+	wget https://github.com/multiarch/qemu-user-static/releases/download/v2.9.1-1/qemu-aarch64-static.tar.gz 
+	tar zxvf qemu-aarch64-static.tar.gz -C bin
+	rm qemu-aarch64-static.tar.gz
+
+build: calico/go-build 
+
+calico/go-build: bin/qemu-ppc64le-static bin/qemu-aarch64-static
 	# Make sure we re-pull the base image to pick up security fixes.
 	docker build --pull -t $(ARCHIMAGE) -f $(DOCKERFILE) .
 
@@ -38,3 +50,5 @@ defaulttarget:
 	docker tag $(ARCHIMAGE) $(DEFAULTIMAGE)
 	docker push $(DEFAULTIMAGE)
 
+clean:
+	rm -rif bin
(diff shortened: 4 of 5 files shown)


djlwilder:
Lets call this an RFC for now :) I added notes into the README on how to do a crossbuild. 

@deitch Is this what you had in mind?

I was able to use this to cross-build a ppc64le felix container on amd64.  However,  it was not that easy, several changes are needed to the felix Makefile.  Most notable adding  -e GOARCH=$(ARCH) when running go-build. We need to use insure the amd64 calico/protoc is used not the ppc or amd arch one.  And finally the Dockerfile-<ppc64le, arm64> must be updated to add qemu-user into the container as a first step or the docker build will fail.  We can remove it later,  but leaving it allows for running the container on the non-native arch,  nice for testing I guess.

deitch:
> Is this what you had in mind?

Yeah. But I wanted to be sure all the parts worked. I hadn't gotten as far as trying felix and other parts yet... glad you did. :-)

Do you have a branch of felix with the changes you needed?

djlwilder:
Two goals with this RFC.
1) Create an amd64 gobuild image that can be used when cross-building go code for other archs.
2) Build go-build images for arm and ppc, building the images on amd64.  Eventual create multi-arch manifest images.

I have had good luck with goal 1,  I can build calico/felix containers for all three arches on an arm64 box.  More on this later.  Unfortunately cross  building go-build containers using qemu is not stable, qemu hangs or crashes.  Using this branch (running on a amd box) I run "ARCH=arm64 make"

deitch:
Referencing [comment](https://github.com/projectcalico/go-build/issues/24#issuecomment-363617794)

> I don't see the need to install packages for the non-native arch into the container. Everything executed should be amd64 amd64 with the exception of binaries we just built and sometimes we need to do that therefore qemu is needed in the go-build:amd64 container.

OK, so we have `Dockerfile.arm64` for building arm64 on arm64, and `Dockerfile.amd64` for building x86 on x86 *or* arm64 on x86 or etc.

Why do we need qemu? Why can we not just do `GOARCH=arm64 go build .` on x86? Of course, when building natively (x86 on x86, arm64 on arm64, etc.) we can just do `go build .`

djlwilder:
I can cross build felix containers from my branch (work in progress :):
git clone https://github.com/djlwilder/felix.git -b cross-build

Using calico/go-build:latest-amd64 built from this PR, I can build all images using the following commands:
make calico/felix   <<<<  builds amd image
BUILDARCH=amd64 ARCH=ppc64le make calico/felix    
BUILDARCH=amd64 ARCH=arm64 make calico/felix

djlwilder:
> Why do we need qemu? Why can we not just do GOARCH=arm64 go build . on x86? Of course, when building natively (x86 on x86, arm64 on arm64, etc.) we can just do go build .

Maybe we don't, I am not sure.  One of the felix build steps "make static-checks" will create a binary using go build then execute it.  I added qemu to handle that,  maybe it could avoided by not cross building in that step.  Other make targets may need it as well IDK.  Qemu must be added to any Dockerfile that will used to an image for a non-native arch.

deitch:
Ah for running the tests. Ok I get it. I’ll dig into it next week. I don’t like that we need qemu static locally for the COPY, but there are ways around it. How did you install qemu static?

djlwilder:
> I don’t like that we need qemu static locally for the COPY, but there are ways around it. How did you install qemu static?

Not sure I understand what you mean...
We can bind mount the qemu into the running container rather than COPY it in.  That works when we are running the amd image to do cross builds.  Qemu is also needed (and must be copied in) when doing a "docker build" of a container for a non-native arch.   That is why I added it into the ppc and amd Dockerfiles.  You'll see that in the felix changes as well.   Support for bind mounts (docker -v option) appear to have been dropped from the docker build command.  Therefore copy is the only option.

> How did you install qemu static?
In the Makefile:
wget https://github.com/multiarch/qemu-user-static/releases/download/v2.9.1-1/qemu-aarch64-static.tar.gz

I also added to the Makefile:  (This loads and sets up binfmt_misc on the host.)
sudo docker run --rm --privileged multiarch/qemu-user-static:register || true

deitch:
So the build process looks something like this. I am trying to play it back, tell me if I missed anything.

## native (BUILDARCH=ARCH)
Nothing special. We don't need qemu, we don't need any changes from what is there. The image is built using `Dockerfile.<ARCH>`. No changes. 

* To make the `go-build` image: `make calico/go-build`
* To push the `go-build` image: `make pusharch`
* To build something dependent on this, e.g. [felix](https://github.com/projectcalico/felix), go into that dir and do the usual build, which is, e.g. `make calico/felix`

All tests will run normally, because the binary (and the image enclosing it) is native to whatever platform we are on.

## cross (BUILDARCH=amd64, ARCH=ppc64le || ARCH=arm64)
There are several considerations:

* Make a cross-build supporting image (this repo)
* Build the actual binary from the cross-build image, e.g. build `felix` for `arm64` while running on `amd64`
* Test the binary, e.g. test `felix` built for `arm64` while running on `amd64`
* Package the binary, e.g. build the image that contains `felix` `arm64` binary and all its dependencies while running on `amd64`

All of the above has to work only in the image generated by `Dockerfile.amd64`, since we only are doing cross-builds from `amd64` to other platforms (and not the other way around). 

Essentially, we don't need to make any changes for the first step (cross-build image) or second step (make the binary), since `go build` can be passed `GOARCH=` env var and it will "just work" (ignoring the missing directories minor change [here](https://github.com/projectcalico/go-build/pull/25/files#diff-c8b09f88c044c3dec6a30981db2282e9R67) ).

Once we have a binary, we need steps 3 (test, which means run the binary) and 4 (package). We can ignore 4 for now, as that is an issue for each of the repos, not this go-build one. That leaves 3. We need a way to run, e.g. an `arm64` binary on `amd64` arch, which qemu is perfect for. But we only ever will do that while on `amd64` (only supported cross-build platform for now). So why do we need to install `qemu` into `Dockerfile.arm64` and `Dockerfile.ppc64le`? 

I don't think that the image generated from `projectcalico/go-build/Dockerfile.arm64`, i.e. `calico/go-build:VERSION-arm64`, would ever be run on any platform other than `arm64`.

## Summary
Trying to summarize, here are my questions:

1. Do we need `qemu` binaries in anything other than the cross-build-supporting image, i.e. `Dockerfile.amd64`?
2. If not, should we just download it directly in the `Dockerfile.amd64` rather than outside in the `Makefile`?
3. Why do we need the `register` target here in the `go-build` repo? Wouldn't it happen only from other repos? Or is it just so we have one central place to demonstrate where/how to use it?

I am going to do another run of full builds now, see what I get.

deitch:
Hmm, I am beginning to see the issue that @tomdee raised [here](https://github.com/projectcalico/go-build/pull/23#issuecomment-362676305) about certain installs.

The `go-build` image contains not just go compilation elements, but also a lot of other packages pre-installed:

* `glibc`
* `std`
* `glide`
* `ginkgo`
* `gometalinter`
* `licenses`
* `gocovmerge`
* `yaml`
* `goveralls`

Some are part of the build itself (e.g. `glibc`), others as tools in the build chain. I see a few options:

* Leave as is. When we do a build with `GOARCH=ppc64le go build .`, it should find and compile as needed ("should"; would have to test). This also will not work for buildchain tools, but since they are running directly on `amd64` only anyways, it shouldn't matter.
* Cross-compile the libs, so they are ready.

I am leaning towards the second. Since we already know the various archs we support (we have to, in order to install the correct `qemu`), why not prepare the libs as well?

deitch:
@djlwilder , how did you get felix to compile with its dependencies as above?

deitch:
And now I discover the dependency on `calico/protoc`, which is compiled for a given arch. It looks like it just generates a protobuf binding file (text), so it does not need to be cross-platform compiled. Is that correct?

deitch:
I have been working on a variant of this. Mainly took your base, then modified as needed.

Check out the branch [cross-compile](https://github.com/deitch/go-build/tree/cross-compile)

I took the qemu out of the arm64 and ppc64le images, keeping them only in amd64 since that is the only place we need them, when cross-compiling. A few other changes as well.

I also have a branch of felix based on this. Compilation and image-building works, deb/rpm are amd64 only, cleaned up the dockerfile naming to keep it consistent, changed the dockerfiles a bit, etc.

binfmt for running the build is a wonderful thing. The challenge is that circleci runs some older kernels. I think 2.0 was running 4.4 (even on Ubuntu 14.04, go figure). Kernels from 4.8+ can use binfmt as long as the qemu binaries exist anywhere on the system (same container, host, another container, doesn't matter), which makes things significantly easier. This is how linuxkit does it.

If I can get that working, we won't need qemu anywhere except in the `go-build` images.

deitch:
> Maybe we don't, I am not sure. One of the felix build steps "make static-checks" will create a binary using go build then execute

I dug into that. I see no reason to run `make static-checks` to run across archs. Linting, checking license versions, etc., not to mention the `ut` targets which look at unit tests and code coverage, etc. These all can happen on the single platform.

The only thing, then, that we would need `qemu` and `binfmt` for would be to build our containers, which are a lot of `RUN apk add` and some file `COPY`/`ADD`. Going to see if there is a better way here...

deitch:
> Going to see if there is a better way here...

Or not. I was hoping to just do `apk --arch <TARGETARCH> -p /out` and use multistage docker builds. It failed for several reasons, but the biggest is that many packages have scripts `preinstall`,`postinstall`,`trigger` - that will run in that root, and therefore need to execute binaries, which means you need a way to run cross-arch, which means qemu and binfmt, which means you might as well just run the native container anyways.

deitch:
OK, I got it all working, without including the qemu in the Makefile or the final image. Took far too long, but it works.

I opened a new PR. @djlwilder , it was based on yours with lots of changes, so feel free to update your PR with my changes (if they meet with approval, of course), and I will close my PR, or vice-versa.

I also opened a felix PR. 

This has been hard... but fun!

djlwilder:
Thanks @deitch,  I will take a look.

djlwilder:
I am closing out this pull request in favour of https://github.com/projectcalico/go-build/pull/26

@deitch you asked a bunch of questions earlier in this PR. I started to answer them but assumed you figured out what you needed in the two new PRs..  let me know if you have more questions.  Thanks for your help.

deitch:
Hey @djlwilder ; I didn't mean you had to close this one. I am more than happy to close mine. I just opened it so we could have a place to compare them. I found some time today and pushed to get some work done on it.

djlwilder:
no worries @deitch,  your PR looks cleaner.  Mine was just an experiment mostly.  

I had added qemu to the amd and ppc go-build images in an attempt to cross build the go-build images themselves.  This would allow a "official" go-build images or a multi-arch image.  Unfortunately, I hit the issues  that  @tomdee raised. "running go under qemu raises exceptions".  At this point I will abandon cross building go-build images.  This only means that anyone doing a native build (other than amd) must build and their own go-build.

PR #26, opened 2018-02-13
cross-compile support

deitch:
Add support **just on amd64** to cross-compile to other supported architectures, notably `arm64` and `ppc64le`. Also add documentation to the `README.md`.

Cross-compiling _from_ other architectures is not supported.

Ideally, we never would need to cross-compile, just build arm64 on arm64, amd64 on amd64, ppc64le on ppc64le, etc. However, with an automated CI toolchain, we need the ability to build all platforms automatically. This does not work since almost no CI-as-a-service offers more than amd64. Thus, we prepare for cross-compile.

This image has been tested with felix, see the other PR referencing this one from felix.

**Note:** I am not wholly convinced that we need `qemu-*-static` in the build container. Since we do not _run_ the cross-compiled image, only _build_ it, it should not matter. However, it is easier with the dependency there.

Fixes: https://github.com/projectcalico/go-build/issues/24

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .idea
+hello-*
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -1,5 +1,8 @@
-FROM golang:1.9.2-alpine3.6
+FROM amd64/golang:1.9.2-alpine3.6
 MAINTAINER Tom Denham <tom@projectcalico.org>
+
+ARG QEMU_VERSION=2.9.1-1
+ARG CROSS_ARCHS="aarch64 ppc64le"
 
 # Install su-exec for use in the entrypoint.sh (so processes run as the right user)
 # Install bash for the entry script (and because it's generally useful)
@@ -10,7 +13,7 @@ MAINTAINER Tom Denham <tom@projectcalico.org>
 # Install wget for fetching glibc
 # Install make for building things
 # Install util-linux for column command (used for output formatting).
-RUN apk add --no-cache su-exec curl bash git openssh mercurial make wget util-linux docker tini
+RUN apk add --no-cache su-exec curl bash git openssh mercurial make wget util-linux docker tini file
 RUN apk upgrade --no-cache
 
 # Disable ssh host key checking
@@ -58,6 +61,16 @@ RUN go get github.com/mikefarah/yaml
 RUN wget https://github.com/fasaxc/goveralls/releases/download/v0.0.1-smc/goveralls && \
     chmod +x goveralls && \
     mv goveralls /usr/bin/
+
+# Enable non-native runs on amd64 architecture hosts
+RUN for i in ${CROSS_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
+RUN chmod +x /usr/bin/qemu-*
+
+# When running cross built binaries run-times will be auto-installed,
+# ensure the install directory is writable by everyone.
+RUN mkdir -m +w -p /usr/local/go/pkg/linux_ppc64le
+RUN mkdir -m +w -p /usr/local/go/pkg/linux_arm64
+
 
 # Ensure that everything under the GOPATH is writable by everyone
 RUN chmod -R 777 $GOPATH
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,26 @@
-ARCH ?= $(shell uname -m)
+BUILDARCH ?= $(shell uname -m)
+ARCH ?= $(BUILDARCH)
 
+ifeq ($(BUILDARCH),aarch64)
+        override BUILDARCH=arm64
+endif
+ifeq ($(BUILDARCH),x86_64)
+        override BUILDARCH=amd64
+endif
 ifeq ($(ARCH),aarch64)
-        ARCH=arm64
+        override ARCH=arm64
 endif
 ifeq ($(ARCH),x86_64)
-        ARCH=amd64
+        override ARCH=amd64
 endif
 
 DOCKERFILE ?= Dockerfile.$(ARCH)
 VERSION ?= latest
 DEFAULTIMAGE ?= calico/go-build:$(VERSION)
 ARCHIMAGE ?= $(DEFAULTIMAGE)-$(ARCH)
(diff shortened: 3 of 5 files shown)


deitch:
Reference #25 . cc @djlwilder @tomdee @caseydavenport

djlwilder:
@deitch,  you commented earlier about needed kernel support for binfmt in the semaphore kernel. Have we tested this?  Maybe we could add a test target here to validate it works.  It could be a simple as building a hello-world.go for each arch then executing them to see if they run.  What do you think?

deitch:
> needed kernel support for binfmt

It had been a while, so I dug back into it. Kernel support for `binfmt` goes back while. I doubt we will have a issues with it in any of the CI.

What _is_ more recent is using qemu outside of the container (chroot). For example, let's say we have our qemu as `/usr/bin/qemu-aarch64-static` on the host, and we want to run an aarch64 binary inside a container. Earlier versions would not work unless we also copied the `qemu-aarch64-static` to the same path inside the container. As of ~4.8 (I could be off by a bit), you can register the binary anywhere (host, another container, etc.), and as long as the kernel can get a handle to it at least once at registration, it will use that, even if it is not in the actual aarch64 container.

That isn't an issue here, since we are copying the qemu binary into the container itself. I did that explicitly to avoid issues with pre-4.8 kernels (like nearly every CI-as-a-service, unfortunately).

deitch:
> It could be a simple as building a hello-world.go for each arch then executing them to see if they run. What do you think?

Sure, sounds like a simple plan.

deitch:
FWIW, docker-for-mac has >4.8 kernel, and includes qemu and binfmt built in, so it always works on d4m.

deitch:
> I like the RUN curl method better

OK. I thought it might be easier to use the package - and I really don't expect to change our qemu binaries almost ever, but I am fine with that.

I wanted to use the `apk` package, but it includes dynamically linked binaries only. There are open issues around making them static, but until they do, not sure I want to bother.

deitch:
> I like the RUN curl method better

Updated.

deitch:
> Have we tested this? Maybe we could add a test target here to validate it works. It could be a simple as building a hello-world.go for each arch then executing them to see if they run.

I added this as a `testcompile` target. I do have a strange issue where `qemu` is not actually executing the target command, it just prints version info and exits. The funny thing is the same binary works fine if I do it on `alpine:3.6` container, but not in the `calico/go-build` container.

I know the install works. This last bit and we can move ahead.

deitch:
Oh well. It was an inherited var. Solved. Pushed out. Back to you.

djlwilder:
@deitch thanks for working on this,   Something got squashed in you Makefile, missing tab on the last line I think. 
I started to play with the test stuff you added but did not get very far.  My concern is that we go to far down the road before we find out that our Ci system has some problem with binfmt.  Maybe I worry to much :)  If you like I can work on that separately in another PR.  Come to think of it once Ci runs once on my PR I will have my answer :)

deitch:
> thanks for working on this
We get this cross-platform, we all win (and I can get it on worksonarm.com, etc.)

> Something got squashed in you Makefile, missing tab on the last line I think.
Sometimes. I love atom.io, other times... Fixing now.

deitch:
> Something got squashed in you Makefile
Fixed.

deitch:
> My concern is that we go to far down the road before we find out that our Ci system has some problem with binfmt

"too far down the road" = with the build process? I cannot change semaphore here (owner needs to change it), but we can add a `make` target to check it, something like `make checkbinfmt`. Can run that before any other builds.

Mind you, I am not sure how that would help. We don't actually use `binfmt` in this repo, only in the client repos.

Let's change it a bit: can you explain how and when we go too far down before knowing, so we can put a checkpoint in the right place?

djlwilder:
Hi Avi
What you have done with testcompile is fine.  My (maybe unfounded) concern is that I have yet to see this dev-ops system run non-native executables using qemu.  It probably can I just need to prove it to myself.  You are correct we don't use qemu when building this repo but we will be depending on it for all the other Calico pieces. This is just a convenient place to try it.  Lets leave this as is for now.  I can push a second PR (after this one is merged) that will force testcompile to be run.  It only needs to be run once, my PR would not need to be merged, its just a test.

djlwilder:
one small change :)  Add "-e LOCAL_USER_ID=$(shell id -u)" to fix a permission denied error from the go build stage.

``` 
testcompile:
-       docker run --rm -e GOARCH=$(ARCH) -w /code -v ${PWD}:/code $(BUILDIMAGE) go build -o hello-$(ARCH) hello.go
+       docker run --rm -e LOCAL_USER_ID=$(shell id -u) -e GOARCH=$(ARCH) -w /code -v ${PWD}:/code $(BUILDIMAGE) go build -o hello-$(ARCH) hello.go
        docker run --rm -v ${PWD}:/code $(BUILDIMAGE) /code/hello-$(ARCH) | grep -q "hello world"```

deitch:
> What you have done with testcompile is fine

Great.

> My (maybe unfounded) concern is that I have yet to see this dev-ops system run non-native executables using qemu. It probably can I just need to prove it to myself.

Better an unfounded concern addressed than something that should have been a concern unaddressed. I like the approach you suggested. :-) 

Truth is, I would think about making `make testcompile` part of the CI process, just so we get an early flag if something is haywire.

deitch:
> one small change :) Add "-e LOCAL_USER_ID=$(shell id -u)" to fix a permission denied error

Done.

What else do we need?

djlwilder:
I ran a test to verify that cross build and qemu work on Semaphore. It ran fine.
My test PR:  [](url) https://github.com/projectcalico/go-build/pull/28
I added commit a340f1b on my fork.  I added a "test:" target that runs testcompile once for each arch.

deitch:
No problem. I can add the same `test` target to this PR. What else do we need to get it in?

deitch:
Doing it slightly differently; I am just going to read arches from "which target-specific Dockerfiles do we have?" That turns the process of creating a new arch from "create new Dockerfile and insert into Makefile" to "create new Dockerfile."

deitch:
> I can add the same test target to this PR. 

Done.

What else?

deitch:
> Wonderful!, I think we are done :)

Excellent, was a bit of a back-and-forth haul, but well worth it. How do we get it merged in?

deitch:
Excellent! Thank you @fasaxc .

PR #27, opened 2018-02-16
Added Dockerfile.s390x to support for s390x

guirish:
Added Dockerfile.s390x file to support for s390x


djlwilder:
Woo-hoo!  Another architecture to add to the mix.  @fasaxc @tomdee  @deitch :)

deitch:
Cool! Looks like we will need to add more `qemu` etc for the `amd64` image.

First need to get it merged in, o' course!

guirish:
@djlwilder , I will update the Dockerfile.s390x for goveralls binary, taking reference of Dockerfile.ppc64le.

guirish:
@fasaxc it picks and updates **BUILDARCH=s390x** as required, on executing "uname -m" command.

PR #31, opened 2018-03-06
Installing qemu-s390x-static in amd image.

djlwilder:
Should fix make test failure.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -68,10 +68,7 @@ RUN chmod +x /usr/bin/qemu-*
 
 # When running cross built binaries run-times will be auto-installed,
 # ensure the install directory is writable by everyone.
-RUN mkdir -m +w -p /usr/local/go/pkg/linux_ppc64le
-RUN mkdir -m +w -p /usr/local/go/pkg/linux_arm64
-RUN mkdir -m +w -p /usr/local/go/pkg/linux_s390x
-
+RUN for arch in ${CROSS_ARCHS}; do mkdir -m +w -p /usr/local/go/pkg/linux_${arch}; done
 
 # Ensure that everything under the GOPATH is writable by everyone
 RUN chmod -R 777 $GOPATH
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -2,7 +2,7 @@ FROM amd64/golang:1.9.2-alpine3.6
 MAINTAINER Tom Denham <tom@projectcalico.org>
 
 ARG QEMU_VERSION=2.9.1-1
-ARG CROSS_ARCHS="aarch64 ppc64le"
+ARG CROSS_ARCHS="aarch64 ppc64le s390x"
 
 # Install su-exec for use in the entrypoint.sh (so processes run as the right user)
 # Install bash for the entry script (and because it's generally useful)
@@ -70,6 +70,7 @@ RUN chmod +x /usr/bin/qemu-*
 # ensure the install directory is writable by everyone.
 RUN mkdir -m +w -p /usr/local/go/pkg/linux_ppc64le
 RUN mkdir -m +w -p /usr/local/go/pkg/linux_arm64
+RUN mkdir -m +w -p /usr/local/go/pkg/linux_s390x
 
 
 # Ensure that everything under the GOPATH is writable by everyone
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,8 @@ DEFAULTIMAGE ?= calico/go-build:$(VERSION)
 ARCHIMAGE ?= $(DEFAULTIMAGE)-$(ARCH)
 BUILDIMAGE ?= $(DEFAULTIMAGE)-$(BUILDARCH)
 
-ARCHES=$(patsubst Dockerfile.%,%,$(wildcard Dockerfile.*))
+# We cross build these arches.
+ARCHES=amd64 arm64 ppc64le
 
 all: build
 


djlwilder:
Unfortunately the s390x test is still failing.  I am not sure if this is a qmeu issue of go issue.
```
# ./hello-s390x
Illegal instruction (core dumped)
```
@guirish   Any ideas?
@deitch - We may need a selective list of architectures that can be cross built, not all-arch.

djlwilder:
I updated the Makefile to only run the cross build test on arm64, and ppc64le. I did not remove s390x-qemu-static from the images however.  Should I?

Dockerfile.amd64
```diff
@@ -70,6 +70,7 @@ RUN chmod +x /usr/bin/qemu-*
 # ensure the install directory is writable by everyone.
 RUN mkdir -m +w -p /usr/local/go/pkg/linux_ppc64le
 RUN mkdir -m +w -p /usr/local/go/pkg/linux_arm64
+RUN mkdir -m +w -p /usr/local/go/pkg/linux_s390x
```
deitch: Do you want to replace these with 

```
RUN for arch in ${CROSS_ARCHS}; do mkdir -m +w -p /usr/local/go/pkg/linux_${arch}; done
```

Save any future pain, and one less manual step to update.

Makefile
```diff
@@ -20,7 +20,8 @@ DEFAULTIMAGE ?= calico/go-build:$(VERSION)
 ARCHIMAGE ?= $(DEFAULTIMAGE)-$(ARCH)
 BUILDIMAGE ?= $(DEFAULTIMAGE)-$(BUILDARCH)
 
-ARCHES=$(patsubst Dockerfile.%,%,$(wildcard Dockerfile.*))
+# We cross build these arches.
+ARCHES=amd64 arm64 ppc64le
```
deitch: Is this because of the comment about "selective list of architectures"?

deitch:
@djlwilder how do I replicate this? Did you just do a `make testcompile ARCH=s390x`? 

It was worth your suggestion to add the `testcompile` target, if only for this. Nice one!

deitch:
Also @djlwilder when you ran this:

```
# ./hello-s390x
Illegal instruction (core dumped)
```

Was it via qemu or on an actual `s390x`?

deitch:
And if qemu, was it properly register with binfmt? I used your `make register`, but you can check if it was registered with `ls /proc/sys/fs/binfmt_misc/` and check if the qemu binary is available?

guirish:
@djlwilder , We have verified the cross-built 'hello-s390x' binary on IBM z systems and its working fine for us. However, its perfectly valid to get an error while executing it on non-s390x architectures as there is difference in machine endian-ness. IBM z is big endian platform so the binary is not supposed to be run on amd64 which is little endian.
IMO test execution should differ based on endian-ness of architectures.

deitch:
Thanks @guirish . That makes sense. It means the build was good but running it under qemu was not. 

> However, its perfectly valid to get an error while executing it on non-s390x architectures as there is difference in machine endian-ness. IBM z is big endian platform so the binary is not supposed to be run on amd64 which is little endian

Shouldn't running it under `qemu` manage that?

I haven't actually run a local vm of a z-system, although I have `qemu-system-s390x` installed locally. If I can get a proper kernel and initrd somewhere, happy to try it. @guirish recommendation? Is there something ready-to-run and easy to install?

cwsolee:
Hi @deitch,  I was trying to offer z VM to Calico for CI purpose earlier but z doesn't support semaphoreCI yet, let me know if there are ways to connect to CI, I could set aside some z resource for CI. For individual use with expiry date, u can get free access from this link https://developer.ibm.com/linuxone/.

deitch:
There are some ways to build them in, but it can slow down. I think the goal now is cross-compiling, if it works.

I will try running the binary, and see if it is a qemu issue vs build issue.

deitch:
OK, requested an account.

deitch:
Hmm...

```
# /usr/bin/qemu-s390x-static -strace /code/hello-s390x
15 sched_getaffinity(0,8192,274886290736,274886299096,274886299344,0) = 16
15 mmap(0x0000004000800bd0,65536,PROT_NONE,MAP_PRIVATE|MAP_ANONYMOUS,-1,0) = 0x000000c000000000
15 munmap(0x000000c000000000,65536) = 0
15 mmap(0x0000004000800ae8,262144,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANONYMOUS,-1,0) = 0x0000004000801000
--- SIGILL {si_signo=SIGILL, si_code=1, si_addr = 0x0000000000068874} ---
Illegal instruction
```

Is this a qemu issue? Or the compiled binary issue?

cwsolee:
U're using no-arch binaries? Likely won't work on z, we've published z specific instruction here https://github.com/linux-on-ibm-z/docs/wiki/Building-Calico while working on all these PR. We submitted 8 PRs recently.

deitch:
> U're using no-arch binaries? 

You lost me?

deitch:
I can confirm (cc @djlwilder ) that this output:

```
# ./hello-s390x
Illegal instruction (core dumped)
```

occurs on qemu. When I run the cross-compiled binary on an actual `s390x` (thanks again @cwsolee for pointing me towards running a LinuxOne vm), it runs fine.

This means it is a qemu issue. I am not sure if we are missing some arguments to qemu, or if there just is an issue.

cwsolee:
Thanks for confirming.

deitch:
If I have time, I will try to track down the qemu issue. No guarantees, though.

guirish:
@deitch I tried following combinations to test and validate QEMU, using the associated binaries built for testing on respective ARCH's:

Test ARCH-->Binary built on-->QEMU used-->Result
x86_64-->x86_64-->qemu-x86_64-static-->passed
x86_64-->s390x-->qemu-s390x-static-->failed
s390x--> x86_64-->qemu-x86_64-static-->failed
s390x-->s390x-->qemu-s390x-static-->failed

I believe the issue is with QEMU w.r.t s390x.

deitch:
We haven't been trying to do all of the cross-builds, just 

* native->native
* amd64->{s390x,aarch64, ... } (cross-building)

The only real reason for cross-building is the prevalence of amd64 as a platform, especially in CI, so we can automate building for everything. As long as `s390x`->`s390x` works, and `amd64`->`s390x` works on the real OS but not on qemu, we know we have some issue, but it is a step forward.

djlwilder:
@deitch, @guirish  thanks for verifying.
I tried upgrading qemu-s390x-static to 2.11.0 but it did not help.

We will need to exclude s390x from cross building until we can solve the problem (here and in the other repos, once we get those done).   @cwsolee this setback will not prevent native builds of Calico for s390x,  I believe the PRs you referenced adds s390 native builds.

deitch:
Correct, @djlwilder . I tried the same. I would love to know why it gives `SIGILL`, though.

@cwsolee or @guirish , do you have that skillset?

cwsolee:
Correct, this PR is for native build. We'll take a look at this too, I could see whether I can find other team in IBM that might have some knowledge on qemu, I personally haven't done much work on this.

deitch:
> I could see whether I can find other team in IBM that might have some knowledge on qemu

Much appreciated.

Makefile
djlwilder: Correct,  we need to remove architectures like s390x that wont cross build.  This list could change over time.  We don't want to remove all support for s390x, just for cross.  As we are just doing a test here it is not all that important.  However, I assume we will have a similar list in other repos as well, that where it will become important.

deitch: Actually, cross-build does work; it is qemu that doesn't. We should figure it out, but I want to leave it for now.

djlwilder: Agreed.

deitch:
I have been trying to launch a full Linux vm using qemu emulating s390x. I figure running the file in there would give us a good sense of the qemu limits and allow some reasonable debugging.

I have had no luck getting it up and running. Using the debian kernel and initrd and `qemu-system-s390x`, along with qemu user networking, but even though the network device is visible and has the right mac address, I cannot get debian to pick up the IP via dhcp.

Anyone had luck with it?

```
qemu-system-s390x -M s390-ccw-virtio -m 1024 -smp 1 -nographic -netdev user,id=mynet0,hostfwd=tcp::10022-:22 -device virtio-net-ccw,netdev=mynet0,id=net0,mac=08:00:2F:00:11:22,devno=fe.0.0001 -kernel ./kernel.debian -initrd ./initrd.debian
```

I basically get a screen offering config of network

```
┌──────────────────┤ [!!] Configure the network device ├──────────────────┐
  │                                                                         │
  │ Please choose the type of your primary network interface that you       │
  │ will need for installing the Debian system (via NFS or HTTP). Only      │
  │ the listed devices are supported.                                       │
  │                                                                         │
  │ Network device type:                                                    │
  │                                                                         │
  │  ctc: Channel to Channel (CTC) or ESCON connection                      │
  │  qeth: OSA-Express in QDIO mode / HiperSockets                          │
  │  iucv: Inter-User Communication Vehicle - available for VM guests       │
  │  virtio: KVM VirtIO                                                     │
  │                                                                         │
  │     <Go Back>                                                           │
  │                                                                         │
  └─────────────────────────────────────────────────────────────────────────┘
```

I figure it wants `virtio`, but doesn't pick up the dhcp address. Anyone done this?

PR #41, opened 2018-04-14
Fix CROSS_ARCHS for golang

deitch:
The commit [here](https://github.com/projectcalico/go-build/commit/883b931c405a7d81541b9ce23b03283a4c07019b#diff-c8b09f88c044c3dec6a30981db2282e9) replaced the explicit `mkdir /usr/local/go/pkg/linux_<arch>` for each arch with the list from `CROSS_ARCHS` in `Dockerfile.amd64`. Unfortunately, `CROSS_ARCHS` was tailored for the `qemu` arch names, which use `aarch64`, and not the golang ones, which use `arm64`.

That merge makes `v0.12` break for all `arm64` builds.

This PR fixes it.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -2,7 +2,13 @@ FROM amd64/golang:1.10.1-alpine3.7
 MAINTAINER Tom Denham <tom@projectcalico.org>
 
 ARG QEMU_VERSION=2.9.1-1
-ARG CROSS_ARCHS="aarch64 ppc64le s390x"
+
+# we need these two distinct lists. The first one is the names used by the qemu distributions
+# these second is the names used by golang see https://github.com/golang/go/blob/master/src/go/build/syslist.go
+# the primary difference as of this writing is that qemu uses aarch64 and golang uses arm64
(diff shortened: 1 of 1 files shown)


Dockerfile.amd64
```diff
@@ -2,7 +2,8 @@ FROM amd64/golang:1.10.1-alpine3.7
 MAINTAINER Tom Denham <tom@projectcalico.org>
 
 ARG QEMU_VERSION=2.9.1-1
-ARG CROSS_ARCHS="aarch64 ppc64le s390x"
+ARG QEMU_ARCHS="aarch64 ppc64le s390x"
```
fasaxc: Please can you comment these two so that we don't make the same mistake again?  Probably at least worth saying that the CROSS_ARCHS one needs to match Go's arch names.

deitch: Yes, quite. Pushing out now.

deitch: Updated. Awaiting your review.

deitch:
Rebased.

PR #42, opened 2018-04-17
Sync packages between Dockerfiles.

djlwilder:
Add the docker and file packages into all Dockerfiles.

The list of packages installed had become inconsistent between architecture. Not all architectures were installing docker and file.  The docker command is used for some unit tests.  File is helpful when working with cross arch images.

### DIFF
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -12,7 +12,7 @@ ARG MANIFEST_TOOL_VERSION=v0.7.0
 # Install wget since it's useful for fetching
 # Install make for building things
 # Install util-linux for column command (used for output formatting).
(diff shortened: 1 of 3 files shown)


djlwilder:
>I don't actually know what we need docker and file for in the go-build images, but if we have it in one, should have it in all.

The test-containerised target in libnetwork-plugin ( and possibility other ut tests) will start a dind container.  The docker binary is needed in go-build to talk to the dockerd in the dind container.  In the past libnetwork-plugin's Makefile added docker into the go-build container for this purpose.  Recently, to simplify this, docker was added directly in the go-build creation ,  but not all Dockerfiles were updated, I have corrected this.  This will allow tests to be run on all supported arch hardware.

The file command can be used to find what arch a binary is build for.  This is very helpful when troubleshooting cross-build issues.

Issue #45, opened 2018-05-17
Cross compiles are slow due to rebuilding stdlib


fasaxc:
Since we do cross compiles using the amd64 image, it'd speed up the downstream builds if we built the standard library for each of the target arches.

tomdee:
This is now fixed

PR #46, opened 2018-05-23
Add pre-built stdlib for cross-compile platforms

deitch:
As pointed out in #45 , cross compiles are slow because it needs to build the stdlib components. This pre-builds it in for all supported cross-compile archs. 

Also adds standard `image` target to `Makefile`.

cc @fasaxc


deitch:
This probably should get a version bump? Not sure how we do that....

fasaxc:
I'll do that once we have a few changes batched up.  Cutting a new version is pretty trivial, but it does need a PR to each repo to rev it so I try to batch up where I can.

PR #48, opened 2018-07-10
Document how to use binfmt for calico projects; fix s390x usage of it

deitch:
We already had all of the `qemu` static binaries installed in our `go-build` environment. This adds the 2 files (script and config) necessary to register support for them via `binfmt`.

Once this is done, you can run:

```
docker run -it --rm --privileged --entrypoint=/usr/local/bin/register calico/go-build:<version>
```

and you then will be able to run binaries for alternate architectures. 

Once this is in (and part of a release), will change the `register` target (which wasn't working) on the other projects that require it for building images.

Fixes what is broken in the cross-image [here](https://github.com/projectcalico/felix/pull/1852)

cc @fasaxc @tomdee @caseydavenport @mkumatag

### DIFF
--- a/Dockerfile.s390x
+++ b/Dockerfile.s390x
@@ -1,7 +1,22 @@
+FROM alpine:3.7 as qemu
+
+ARG QEMU_VERSION=2.9.1-1
+ARG QEMU_ARCHS="s390x"
+
+RUN apk --update add curl
+
+# Enable non-native runs on amd64 architecture hosts
+RUN for i in ${QEMU_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
+RUN chmod +x /usr/bin/qemu-*
+
 FROM s390x/golang:1.10.1-alpine3.7
 MAINTAINER LoZ Open Source Ecosystem (https://www.ibm.com/developerworks/community/groups/community/lozopensource)
 
 ARG MANIFEST_TOOL_VERSION=v0.7.0
+
+# Enable non-native builds of this image on an amd64 hosts.
+# This must be the first RUN command in this file!
+COPY --from=qemu /usr/bin/qemu-*-static /usr/bin/
 
 # Install su-exec for use in the entrypoint.sh (so processes run as the right user)
 # Install bash for the entry script (and because it's generally useful)
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,12 +1,12 @@
 FROM alpine:3.7 as qemu
 
 ARG QEMU_VERSION=2.9.1-1
-ARG CROSS_ARCHS="aarch64 ppc64le"
+ARG QEMU_ARCHS="aarch64"
 
 RUN apk --update add curl
 
 # Enable non-native runs on amd64 architecture hosts
-RUN for i in ${CROSS_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
+RUN for i in ${QEMU_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
 RUN chmod +x /usr/bin/qemu-*
 
 FROM arm64v8/golang:1.10.1-alpine3.7
--- a/Dockerfile.ppc64le
+++ b/Dockerfile.ppc64le
@@ -1,12 +1,12 @@
 FROM alpine:3.7 as qemu
 
 ARG QEMU_VERSION=2.9.1-1
-ARG CROSS_ARCHS="aarch64 ppc64le"
+ARG QEMU_ARCHS="ppc64le"
 
 RUN apk --update add curl
 
 # Enable non-native runs on amd64 architecture hosts
-RUN for i in ${CROSS_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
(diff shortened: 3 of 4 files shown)


mkumatag:
But my basic question is - why we are trying to run the other arch binary in the `go-build` container in the first place? what is the basic need for this?

Via `go-build` we should be just building the target arch binary and publish. Sometimes even after the registering the, there is a chance that other arch binaries may not run just like that(it might fail with some missing symbol).

IMO, The intention is good for adding the register and all fine, but we shouldn't depend on running any cross-built binaries on a different architecture.(e.g here running arm64 binaries on amd64 arch)

deitch:
> why we are trying to run the other arch binary in the go-build container in the first place?

We aren't. The problem is as follows. When you run the actual registration, the `qemu-*-static` binaries must exist in one of two places:

* the same container as the place you are running other-arch binaries
* the same container as you run the registration process

For the first option, that means adding `qemu-*-static` to _each and every_ `Dockerfile` that does a cross-image. That makes it more complex, and makes the final image bigger (unless we do some even more complicated multistage).

For the second option, it only has to exist once. If it was in the container that did the registration, then any other container (on the same kernel) can find it.

Try it yourself. Download this PR, built it, then run:

```
docker run -it --rm --privileged --entrypoint=/usr/local/bin/register calico/go-build:<version>
```

And then run 

```
docker run -it --rm s390x/alpine:3.7 sh
```

You will get a functioning `sh` that was built for `s390x`, despite never having placed `qemu-s390x-static` in the new `alpine:3.7` container you just ran.

fasaxc:
@mkumatag Using QEMU has allowed us to cross-build docker images from AMD64 so that we can do so from our CI pipeline without having to add additional hardware.

@deitch Change is looking good; couple of questions:

- Would it be worth submitting a PR to the multiarch/qemu repo to allow their docker image to set the F flag.  Feels like that'd be a great feature there.
- Any issues with docker now that the kernel will be keeping that file open after the container finishes?  Feels like docker might struggle to reap the container.

deitch:
I have simplified this down. The kernels we are running - mainly 4.4 in Semaphore for CI - do not support the "interpreter at registration time" flag (that requires 4.8+).

Instead, I am just making this a change in documentation to explain how to use it, and adding some missing pieces from the `s390x` Dockerfile.

Dockerfile.s390x
```diff
@@ -1,8 +1,23 @@
+FROM alpine:3.7 as qemu
+
+ARG QEMU_VERSION=2.9.1-1
+ARG CROSS_ARCHS="aarch64 ppc64le"
```
tomdee: Shouldn't s390x be in this list (and for the other Dockerfiles?)

deitch: Hmm, you are right, this part is wrong.

We actually shouldn't have the list in each, just the arch for itself. We don't build `ppc64le` on `s390x` or `amd64` on `arm64`. We _only_ build each on its own (`s390x` on `s390x`, `arm64` on `arm64`, etc.), or _all_ on `amd64`. So `Dockerfile.amd64` should have all, and `Dockerfile.<anything other than amd64>` should have just itself.

Will fix immediately.

tomdee:
@deitch I'm still a bit confused with what you're trying to achieve with this PR. The documentation part I get, but I'm not sure about the Dockerfile changes. 

IIUC, the near term goal is to get everything building on an AMD64 host, in which case the AMD64 go-build image will be used. The AMD64 dockerfile already has the `qemu-aarch64-static  qemu-ppc64le-static  qemu-s390x-static` files. 

Any changes to the non-amd64 gobuild files are only useful for building on hosts of those architectures, in which case what's the point in just putting the qemu-ARCH-static matching the arch of the Dockerfile in the image? If you want to enable cross-building from a non-amd64 host then shouldn't you put all architectures except for the current one?

deitch:
> The documentation part I get

Heh, good. If the documentation wasn't clear, nothing else would be! :-)

> IIUC, the near term goal is to get everything building on an AMD64 host, in which case the AMD64 go-build image will be used

Almost. We actually _do_ build the `go-build` images for others on `amd64`, so you can use `calico/go-build:<version>-arm64` on an `arm64` host without having to build the entire `go-build` and then build whatever binary you are after. 

Here are the current images for `v0.16:

* `v0.16`
* `v0.16-ppc64le`
* `v0.16-arm64`
* `v0.16-amd64`

The changes to the Dockerfiles do the following:

1. Eliminate unnecessary qemu files (not much point in having a `qemu` file to run `ppc64le` binaries on `amd64` hosts in an `arm64` image!)
2. Add the above to `s390x`. It still doesn't work, for `qemu` issues, which is why it is excepted out.

So it was just eliminating excess and aligning `Dockerfile.s390x`.

Does that answer?

tomdee:
Thanks for the explanation. I'm clearer now on the fact that we want to build the non-amd64 go-build images on an amd64 host. But I'm still a little unclear on what the non-amd64 go-build images will be used for? Is it just building non-amd64 thing binaries natively from non-amd64 hosts? Or is to cross-building from non-amd64 hosts.

If it's the former, then why have any qemu stuff at all in the non-amd64 go-build images?
If it's the latter, then don't we need qemu stuff for_all_ of the supported arches in the non-amd64 go-build images?

I guess Im just struggling to see any point in _just_ having the qemu binary matching the Dockerfile architecture...

deitch:
>  I'm clearer now on the fact that we want to build the non-amd64 go-build images on an amd64 host

Yep, that's it. Make it part of the pipeline.

> Is it just building non-amd64 thing binaries natively from non-amd64 hosts?

Yes. No one has any desire to start supporting, even unofficially, the entire matrix of cross-builds.

> If it's the former, then why have any qemu stuff at all in the non-amd64 go-build images?

Because if you want to build, e.g. `go-build:<version>-arm64` on an `amd64` (which is what "the former" is), then you need to be running all sorts of `arm64` binaries (in the second stage of the `Dockerfile.arm64`), likewise for the other archs. You cannot do that on `amd64` with `binfmt` unless you have the `qemu` interpreter ready in the container (or in the register container for kernel >= 4.8, but none of the CI as a service has that).

To summarize:

* `Dockerfile.amd64` needs all of the qemu interpreters so that later stages (`felix`, `typha`, etc.) have it in a standard location to cross-build non-`amd64` images on `amd64`
* `Dockerfile.<anything_else>` needs the qemu just for that arch so that `go-build` itself can be built.

How's that?

tomdee:
Phew, I was just being a little slow. Thanks for your patience, I've got it clear in my head now!

deitch:
> Phew, I was just being a little slow

Not at all. I wrote this darn PR, and don't ask how many times I had to reedit it to make it make sense. 

Thanks for looking and merging.

PR #59, opened 2018-10-04
Add docker for use in some STs

mgleung:
docker is required for properly running the STs in some repos.


tomdee:
How much bigger does this make the image?

mgleung:
~170 MB

PR #60, opened 2018-10-24
Add go-junit-report tool

lwr20:
This tool is needed to convert go-test output into junit test reports (for consumption by a test management system).  

This has not been needed up to now because Ginkgo provides a junit reporter natively (and our repos used Ginkgo for tests), but with the rise of repos which do not use Ginkgo for tests (e.g. app-policy), this tool is now needed.


lwr20:
License on `github.com/jstemmer/go-junit-report` is MIT.

Issue #64, opened 2019-01-25
Can't use a big UID


namreg:
Hi there.

I have a big user id (1069552525) in my ubuntu system. (the uid was assigned by my system administrator). When I try to use `go-build` image with passing a `LOCAL_USER_ID` env variable, I get a warning message (the second line):
```
Starting with UID : 1069552525
adduser: number 1069552525 is not in 0..256000 range
```
The case in that alpine `adduser` command has a limit on user id.

The case when it is critical for me:
I try to build the felix project (https://github.com/projectcalico/felix). When I run `make build` LOCAL_USER_ID=$(shell id -u) is passing to the `go-builder` (https://github.com/projectcalico/felix/blob/master/Makefile#L215), but this uid is not acceptable for the alpine `adduser` command and it fallbacks to the 0 uid (root user, as we know). As a result, a vendor directory creates on behalf of the root user and this step https://github.com/projectcalico/felix/blob/master/Makefile#L285 is failed with permission denied error.

I have linked PR that fixes this problem.

Thanks.

caseydavenport:
Looks the same as https://github.com/projectcalico/calicoctl/issues/1961

PR #65, opened 2019-01-25
Replace adduser with useradd

namreg:
This PR replace alpine's `adduser` command with the `useradd` command from `shadow` package.
The latest does not have a limit on UID. See details in the issue.

Fixes https://github.com/projectcalico/go-build/issues/64


namreg:
@caseydavenport Any thoughts ? :)

PR #69, opened 2019-06-13
Adding /home/user

asincu:
Creating /home/user when creating the user. If /home/user is already created will skip this step.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -9,7 +9,13 @@ USER_ID=${LOCAL_USER_ID:-9001}
 echo "Starting with UID : $USER_ID" 1>&2
 # Do not create mail box.
 /bin/sed -i 's/^CREATE_MAIL_SPOOL=yes/CREATE_MAIL_SPOOL=no/' /etc/default/useradd
-/usr/sbin/useradd -U -s /bin/bash -u $USER_ID user
+# Don't pass "-m" to useradd if the home directory already exists (which can occur if it was volume mounted in) otherwise it will fail.
+if [[ ! -d "/home/user" ]]; then
+    /usr/sbin/useradd -m -U -s /bin/bash -u $USER_ID user
+else
+    /usr/sbin/useradd -U -s /bin/bash -u $USER_ID user
+fi
+
 export HOME=/home/user
 
 if [ -n "$EXTRA_GROUP_ID"  ]; then


entrypoint.sh
```diff
@@ -9,7 +9,10 @@ USER_ID=${LOCAL_USER_ID:-9001}
 echo "Starting with UID : $USER_ID" 1>&2
 # Do not create mail box.
 /bin/sed -i 's/^CREATE_MAIL_SPOOL=yes/CREATE_MAIL_SPOOL=no/' /etc/default/useradd
-/usr/sbin/useradd -U -s /bin/bash -u $USER_ID user
+if [[ ! -d "/home/user" ]]; then
+    /usr/sbin/useradd -m -U -s /bin/bash -u $USER_ID user
+fi
```
ozdanborne:
```suggestion
else
    /usr/sbin/useradd -U -s /bin/bash -u $USER_ID user
fi
```

entrypoint.sh
```diff
@@ -9,7 +9,10 @@ USER_ID=${LOCAL_USER_ID:-9001}
 echo "Starting with UID : $USER_ID" 1>&2
 # Do not create mail box.
 /bin/sed -i 's/^CREATE_MAIL_SPOOL=yes/CREATE_MAIL_SPOOL=no/' /etc/default/useradd
-/usr/sbin/useradd -U -s /bin/bash -u $USER_ID user
+if [[ ! -d "/home/user" ]]; then
```
ozdanborne:
```suggestion
# Don't pass "-m" to useradd if the home directory already exists, which can occur if it was volume mounted in, otherwise it will fail.
if [[ ! -d "/home/user" ]]; then
```

entrypoint.sh
asincu: @ozdanborne I added the else branch that adds the user without the home directory. I also added to change the ownership of /home/user to the current user.

entrypoint.sh
```diff
@@ -9,7 +9,14 @@ USER_ID=${LOCAL_USER_ID:-9001}
 echo "Starting with UID : $USER_ID" 1>&2
 # Do not create mail box.
 /bin/sed -i 's/^CREATE_MAIL_SPOOL=yes/CREATE_MAIL_SPOOL=no/' /etc/default/useradd
-/usr/sbin/useradd -U -s /bin/bash -u $USER_ID user
+# Don't pass "-m" to useradd if the home directory already exists (which can occur if it was volume mounted in) otherwise it will fail.
+if [[ ! -d "/home/user" ]]; then
+    /usr/sbin/useradd -m -U -s /bin/bash -u $USER_ID user
+else
+    /usr/sbin/useradd -U -s /bin/bash -u $USER_ID user
+    chown -R $USER_ID /home/user
```
ozdanborne: I don't know if this is going to fly, because it will modify permission of files on the host.

I think a better plan is to stop volume mounting into `/home/user` from the Makefile

PR #73, opened 2019-08-14
Update to golang 1.12.8

tmjd:


Dockerfile.s390x
```diff
@@ -9,7 +9,7 @@ RUN apk --update add curl
 RUN for i in ${QEMU_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
 RUN chmod +x /usr/bin/qemu-*
 
-FROM s390x/golang:1.11.9-alpine3.8
+FROM s390x/golang:1.12.8-alpine3.8
```
lmm: This image doesn't exist but `s390x/golang:1.12.8-alpine3.9` does. I'm not sure if anything needs to be done to use 3.9 over 3.8.

tmjd: Good catch

PR #77, opened 2019-09-02
restore docker binary

bcreane:
Un-do: https://github.com/projectcalico/go-build/commit/4e89359bc716d3efff7efe36c4b57b4c52afa7bf#diff-c8b09f88c044c3dec6a30981db2282e9 since `docker` binary is used in a tigera/private repo: `calicoq`.


bcreane:
Discussed with @caseydavenport - we'll rearrange our repo's STs to not require docker binary in the base go-build image.

PR #78, opened 2019-09-03
Downgrade from golangci-lint to gometalinter

rafaelvanoni:


bcreane:
This will break about 10 TSEE repos at this point. What about including both linters?

PR #79, opened 2019-09-15
s/1.12.8-alpine3.9/1.12.9-alpine3.10/g

### LINKED ISSUE #74: s390x Dockerfile needs base image updated
Most of our Dockerfiles have migrated over to Alpine v3.9 and Go 1.12 but s390x is still using `s390x/golang:1.11.9-alpine3.8`, which means the image is missing some security patches (e.g. [this](https://groups.google.com/forum/#!topic/golang-nuts/fCQWxqxP8aA) most recent one).

Any s390x users want to take this on? 😄

chenrui333:
Bump the version to the latest go v1.12

Fixes #74


chenrui333:
@tmjd, I think we can probably merge this one before moving to go-1.13. I will submit a different one separately.

PR #81, opened 2019-09-20
Add Git config for accessing private repos over SSH

neiljerram:


neiljerram:
I'll close this for now.  It's only a one-liner, and we're currently achieving the same objective with the `GIT_CONFIG_SSH` in `Makefile.common`.

PR #85, opened 2019-10-15
Update to golangcli-lint 1.20.0, because...

neiljerram:
1. Release notes include `95ec0cf dramatically reduce memory
usage (#758)`

2. Hoping it will fix this false positive:

     pkg/log/log.go:20:5: var `tag` is unused (unused)
     var tag string
         ^


neiljerram:
Actually the "false positive" wasn't false, and I've now removed the unused code at https://github.com/tigera/confd-private/pull/68.

neiljerram:
@tmjd Thanks.  Yes, I agree about the workflow and will adjust the branch point accordingly.

PR #87, opened 2019-11-11
Move common Makefile targets to Makefile.common

rafaelvanoni:


Makefile.common
```diff
@@ -0,0 +1,320 @@
...
+		fi'
+endef
+
+TYPHA_BRANCH?=$(PIN_BRANCH)
+TYPHA_REPO?=github.com/projectcalico/typha
+LIBCALICO_BRANCH?=$(PIN_BRANCH)
+LIBCALICO_REPO?=github.com/projectcalico/libcalico-go
+CONFD_BRANCH?=$(PIN_BRANCH)
+CONFD_REPO?=github.com/projectcalico/confd
+FELIX_BRANCH?=$(PIN_BRANCH)
```
tmjd: Should we have pin update targets for CONFD and FELIX here also?

rafaelvanoni: See https://github.com/projectcalico/felix/pull/2178 for an example use case.

tmjd:
Did you consider adding this in a different repo (or even a new one)?
I feel like this doesn't really fit here. What about having it in libcalico-go? Or we could make a new repo but that seems a little silly, but I could be convinced.
Has this discussion already been had and I'm late to the party?

rafaelvanoni:
> Did you consider adding this in a different repo (or even a new one)?
> I feel like this doesn't really fit here. What about having it in libcalico-go? Or we could make a new repo but that seems a little silly, but I could be convinced.
> Has this discussion already been had and I'm late to the party?

There was a short discussion in slack of where to put it. My original idea was to have this in its own repo, but Shaun pointed out we need to a better versioning scheme than branches/releases. So I thought of moving it here and version it along with GO_BUILD_VER.

Makefile.common
tmjd: Thanks for the example. Are you going to add `update-confd-pin` and `update-felix-pin` targets?

Makefile.common
```diff
@@ -0,0 +1,320 @@
...
+endif
+
+# For building, we use the go-build image for the *host* architecture, even if the target is different
+# the one for the host should contain all the necessary cross-compilation tools
+# we do not need to use the arch since go-build:v0.15 now is multi-arch manifest
+CALICO_BUILD=calico/go-build:$(GO_BUILD_VER)
+ETCD_VERSION?=v3.3.7
+K8S_VERSION?=v1.14.1
+PROTOC_VER?=v0.1
+PROTOC_CONTAINER ?=calico/protoc:$(PROTOC_VER)-$(BUILDARCH)
```
tmjd: The code here is correct but in the v0.27 version these 2 lines are mixed up.

rafaelvanoni:
Already fixed

PR #90, opened 2019-12-13
Consolidate build release targets

realgaurav:
Proposing to move more common targets under Makefile.common including build and release targets.
Also updating golang version to _preferred_ version.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -1,4 +1,4 @@
-FROM amd64/golang:1.12.14-alpine3.10
+FROM amd64/golang:1.12.8-alpine3.9
 MAINTAINER Tom Denham <tom@projectcalico.org>
 
 ARG QEMU_VERSION=2.9.1-1
--- a/Makefile.common
+++ b/Makefile.common
@@ -288,6 +288,68 @@ git-push:
 commit-pin-updates: update-pins build git-status git-config git-commit ci git-push
 
 ###############################################################################
+# Image build/push
+###############################################################################
+# we want to be able to run the same recipe on multiple targets keyed on the image name
+# to do that, we would use the entire image name, e.g. calico/node:abcdefg, as the stem, or '%', in the target
+# however, make does **not** allow the usage of invalid filename characters - like / and : - in a stem, and thus errors out
+# to get around that, we "escape" those characters by converting all : to --- and all / to ___ , so that we can use them
+# in the target, we then unescape them back
+escapefs = $(subst :,---,$(subst /,___,$(1)))
+unescapefs = $(subst ---,:,$(subst ___,/,$(1)))
+
+# these macros create a list of valid architectures for pushing manifests
+space :=
+space +=
+comma := ,
+prefix_linux = $(addprefix linux/,$(strip $1))
+join_platforms = $(subst $(space),$(comma),$(call prefix_linux,$(strip $1)))
+
+imagetag:
+ifndef IMAGETAG
+	$(error IMAGETAG is undefined - run using make <target> IMAGETAG=X.Y.Z)
+endif
+
+## push one arch
+push: imagetag $(addprefix sub-single-push-,$(call escapefs,$(PUSH_IMAGES)))
+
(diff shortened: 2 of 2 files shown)


Dockerfile.amd64
```diff
@@ -1,4 +1,4 @@
-FROM amd64/golang:1.12.14-alpine3.10
+FROM amd64/golang:1.12.8-alpine3.9
```
caseydavenport: Could you give some detail on this change? What's the reasoning for this?

Makefile.common
```diff
@@ -353,3 +415,88 @@ help:
...
+ifneq ($(VERSION), $(GIT_VERSION))
+	$(error Attempt to build $(VERSION) from $(GIT_VERSION))
+endif
+
+	$(MAKE) image
+	$(MAKE) tag-images IMAGETAG=$(VERSION)
+
+## Verifies the release artifacts produces by `make release-build` are correct.
+release-verify: release-prereqs
+	# TODO: Check the reported version is correct for each release artifact. Uncomment when binary supports version command.
```
caseydavenport: This is all commented out - how do you propose we re-enable this check? 

One challenge is that each repo has its own version command, so we need a way for the calling Makefile to specify the command to run to get the binary version.

Dockerfile.amd64
realgaurav: @rafaelvanoni can shed some light. I was asked to use this go-version.

Makefile.common
```diff
@@ -353,3 +415,88 @@ help:
...
+	# fi
+
+## Generates release notes based on commits in this version.
+release-notes: release-prereqs
+	mkdir -p dist
+	echo "# Changelog" > release-notes-$(VERSION)
+	sh -c "git cherry -v $(PREVIOUS_RELEASE) | cut '-d ' -f 2- | sed 's/^/- /' >> release-notes-$(VERSION)"
+
+## Pushes a github release and release artifacts produced by `make release-build`.
+release-publish: release-prereqs
```
caseydavenport: This varies based on repository as well. I don't think this really make sense to have in the common Makefile.

The entire premise of the release infrastructure is that each repository knows how to build and publish itself. Moving it to this file removes the necessary context required to properly do that. 

e.g., the CNI repository publishes binaries to GitHub, but app-policy does not. How do we represent that here?

Makefile.common
```diff
@@ -353,3 +415,88 @@ help:
...
+	@echo ""
+
+# WARNING: Only run this target if this release is the latest stable release. Do NOT
+# run this target for alpha / beta / release candidate builds, or patches to earlier Calico versions.
+## Pushes `latest` release images. WARNING: Only run this for latest stable releases.
+release-publish-latest: release-prereqs
+	$(MAKE) push-all push-non-manifests IMAGETAG=latest
+
+# release-prereqs checks that the environment is configured properly to create a release.
+release-prereqs:
```
caseydavenport: If you look across the core repos I think you'll spot that this target also varies a little bit.

Makefile.common
```diff
@@ -287,6 +287,68 @@ git-push:
...
+	docker tag $(BUILD_IMAGE):latest-$(ARCH) $(call unescapefs,$*:$(IMAGETAG))
+else
+	$(NOECHO) $(NOOP)
+endif
+
+## tag images of all archs
+tag-images-all: imagetag $(addprefix sub-tag-images-,$(VALIDARCHES))
+sub-tag-images-%:
+	$(MAKE) tag-images ARCH=$* IMAGETAG=$(IMAGETAG)
+
```
doublek: @caseydavenport WDYT about just the targets highlighted by this comment move into the common makefile? I agree that the release ones are repo specific and require further discussion.

caseydavenport: I think that's good to do so long as the IMAGETAG / BUILD_IMAGE / PUSH_IMAGES / etc, etc, can be properly set in the main Makefile and interpreted in this one.

caseydavenport:
Closing since this is fairly old and inactive.

PR #93, opened 2020-01-02
get_remote_version() needs to run in a docker

rafaelvanoni:


caseydavenport:
@rafaelvanoni could you provide a bit more context? Why does this command need to run in a container?

rafaelvanoni:
> @rafaelvanoni could you provide a bit more context? Why does this command need to run in a container?

We need the `$(GIT_CONFIG_SSH)` bit for private repos.

Makefile.common
```diff
@@ -218,15 +218,15 @@ DOCKER_GO_BUILD := $(DOCKER_RUN) $(CALICO_BUILD)
...
+	$(shell $(DOCKER_RUN) $(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) git ls-remote https://$(1) $(2) | cut -f1')
 endef
 
 # update_pin updates the given package's version to the latest available in the specified repo and branch.
 # $(1) should be the name of the package, $(2) and $(3) the repository and branch from which to update it.
 define update_pin
 	$(eval new_ver := $(call get_remote_version,$(2),$(3)))
 
-	$(DOCKER_RUN) $(CALICO_BUILD) sh -c '\
+	$(DOCKER_RUN) -i $(CALICO_BUILD) sh -c '\
```
caseydavenport: What's the `-i` needed for here?

rafaelvanoni: So we can `ctrl-c` the call if needed.

PR #98, opened 2020-02-27
Avoid updating global git config for pin update commit

realgaurav:
## Description

Avoid updating global git config for pin update commit.

I have noticed `git-config` updating global git config. I understand `CONFIRM` is supposed to guard against unintended config update, given the impact I am proposing to avoid global config update to limit the commit info limited to `git-commit`


realgaurav:
@bcreane can you please merge this? Thanks!

PR #100, opened 2020-03-09
Build for 32-bit arm v7

Josua-SR:
This PR introduces the v7 variant of the arm architecture as build target.

Build-tested natively on debian 10 - armhf, cross-compiled on amd64, pushed to https://quay.io/repository/josua-sr/calico_go-build?tab=tags for verifying the manifest.

The second commit has some potentially relevant changes - I tried to make it in a way that no existing names or image tags change - and only added docker platform `arm/v7` with the image tag suffix `-armv7`.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -8,8 +8,8 @@ ARG QEMU_VERSION=4.2.0-6
 # we need these two distinct lists. The first one is the names used by the qemu distributions
 # these second is the names used by golang see https://github.com/golang/go/blob/master/src/go/build/syslist.go
 # the primary difference as of this writing is that qemu uses aarch64 and golang uses arm64
-ARG QEMU_ARCHS="aarch64 ppc64le s390x"
-ARG CROSS_ARCHS="arm64 ppc64le s390x"
+ARG QEMU_ARCHS="arm aarch64 ppc64le s390x"
+ARG CROSS_ARCHS="arm arm64 ppc64le s390x"
 
 ARG MANIFEST_TOOL_VERSION=v1.0.2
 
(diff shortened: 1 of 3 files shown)


fasaxc:
Looks like this failed CI (and we can't accept PRs unless you're able to sign the CLA).  Please can you take another look?

Josua-SR:
> Looks like this failed CI (and we can't accept PRs unless you're able to sign the CLA). Please can you take another look?

afair the CI back then failed generally, not on my particular changes. Either way I did a quick rebase now (untested) - let's see how it turns out!

Josua-SR:
```
calicoctl/commands/ipam/show.go:75:13: assignment mismatch: 2 variables but ipamClient.GetAssignmentAttributes returns 3 values
note: module requires Go 1.13
github.com/projectcalico/calicoctl/calicoctl/commands/node
Makefile:89: recipe for target 'build-calicoctl' failed
make[1]: *** [build-calicoctl] Error 2
```

It seems this is how the CI fails - in calicoctl. My changes should  not impact go-build functionally - and tehrefore neither calicoctl (unless I overlooked something).

Josua-SR:
Last force-push worked around the following issue that I would ask for some advice about:
```
make: *** No rule to make target  "sub-image-arm/v7"
```

The cause is quite simple:
```
image-all: $(addprefix sub-image-,$(ARCHES))
sub-image-%:
	$(MAKE) image ARCH=$*
```

The '%' pattern does **not** match '/' - so in essence for keeping the Makefile short, ARCHES has to name all architectures without a '/ '.
Now this is where advice is needed: What value is supposed to be used in the ARCHES variable? The `uname -m` name of the architecture - or the slang term of the architecture, e.g. arm, armel, armhf - or the docker os/arch/variant format? I took the quick route here and used the first - because it contains no / and was therefore the quickest shortcut.

fasaxc:
Pushed your branch to the main repo as `armv7`, that'll trigger a test build that will push the real container image

fasaxc:
Someone else picked this up under #217

PR #107, opened 2020-05-06
Use EXTRA_COMMIT_PATTERNS env for git-commit target

Brian-McM:
This will allow projects to specify extra files to commit when doing auto pin updates, i.e. generated files

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -299,7 +299,7 @@ ifdef CONFIRM
 endif
 
 git-commit:
-	git diff --quiet HEAD || git commit -m "Semaphore Automatic Update" go.mod go.sum
+	git diff --quiet HEAD || git commit -m "Semaphore Automatic Update" go.mod go.sum $(EXTRA_FILES_TO_COMMIT)
 
 git-push:
 	git push


Makefile.common
```diff
@@ -299,7 +299,7 @@ ifdef CONFIRM
 endif
 
 git-commit:
-	git diff --quiet HEAD || git commit -m "Semaphore Automatic Update" go.mod go.sum
+	git diff --quiet HEAD || git commit -m "Semaphore Automatic Update" go.mod go.sum ${EXTRA_FILES_TO_COMMIT}
```
neiljerram, line 302: I think it would be better to use a Make variable better `$(...)` instead of an environment variable `${...}`. Then it will work to set this either in the environment or in Makefile.

In fact, not sure it works at all like this; I think the `$` would need to be quoted.

Brian-McM: I tested this and it works (set the variable in the apiserver makefile and ran the target), it doesn't need to be quoted, but maybe it should. Quoting is recommend when the variable can contain spaces or be empty, and it's a safety precaution to splitting up the words in the variable.

I'll test out changing the variable to $(...), but I looked at this http://aggregate.org/rfisher/Tutorials/Make/make5.html and it seems seems ${...} is used to access variables as well.

Brian-McM: Update with the suggested changes

Makefile.common
```diff
@@ -299,7 +299,7 @@ ifdef CONFIRM
 endif
 
 git-commit:
-	git diff --quiet HEAD || git commit -m "Semaphore Automatic Update" go.mod go.sum
+	git diff --quiet HEAD || git commit -m "Semaphore Automatic Update" go.mod go.sum "$(EXTRA_FILES_TO_COMMIT)"
```
neiljerram, line 302: I don't think there should be quotes around `$(EXTRA_FILES_TO_COMMIT)`, because `git commit` will then see that as a single arg, and probably treat it as a single file name with strange white space.

Brian-McM: Oh I thought you said you wanted it quoted? I guess I misunderstood what you wanted quoted...

Brian-McM: OK, I've undone that change

PR #112, opened 2020-05-21
Pin to version of k8s without CRD bug

caseydavenport:
Specifically, this one: https://github.com/kubernetes/kubernetes/issues/90904


liggitt:
FYI, https://github.com/kubernetes/kubernetes/issues/90904 is resolved in the latest 1.16/1.17/1.18 patch releases (1.18.4+)

PR #130, opened 2020-08-28
Makefile targets for auto pin update PRs

Brian-McM:
This includes macros to help with communicating with the github API.

In semaphore you would just create a pipeline that runs the target `CONFIRM=true GITHUB_TOKEN=${MARVIN_GITHUB_TOKEN} make trigger-auto-pin-update-process`

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -242,7 +242,7 @@ define update_pin
 	$(DOCKER_RUN) -i $(CALICO_BUILD) sh -c '\
 		if [ ! -z "$(new_ver)" ]; then \
 			$(GIT_CONFIG_SSH) \
-			go get $(1)@$(new_ver); \
+			go get -d $(1)@$(new_ver); \
 			go mod tidy; \
 		fi'
 endef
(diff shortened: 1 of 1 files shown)


Makefile.common
```diff
@@ -311,10 +311,111 @@ git-commit:
 git-push:
 	git push
 
+commit-and-push-pr:
```
doublek, line 314: It could get confusing with this target alongside the other two targets on top which are separate commit and push targets. To elaborate on my concern, normally if I were reading this, I would think that `commit-and-push-pr` was a  just a combination of `git-commit` and a `git-push` but they serve different purposes. Anyway we can combine them reasonably?

There is also a bunch of variables that can all cause confusion. For instance `GIT_COMMIT_EXTRA_FILES` versus `EXTRA_FILES_TO_COMMIT`. I think there could be some cleanup around these as well.

Makefile.common
```diff
@@ -311,10 +311,111 @@ git-commit:
...
+GIT_COMMIT_FILES?=$(GIT_PIN_UPDATE_COMMIT_FILES) $(GIT_PIN_UPDATE_COMMIT_EXTRA_FILES)
+
+# Call the github API. $(1) is the http method type for the https request, $(2) is the repo slug, and is $(3) is for json
+# data (if omitted then no data is set for the request). If GITHUB_API_EXIT_ON_FAILURE is set then the macro exits with 1
+# on failure. On success, the ENV variable GITHUB_API_RESPONSE will contain the response from github
+define github_call_api
+	$(eval CMD := curl -f -X$(1) \
+						-H "Content-Type: application/json"\
+						-H "Authorization: token ${GITHUB_TOKEN}"\
+						https://api.github.com/repos/$(2) $(if $(3),--data '$(3)',))
```
doublek, line 338: nit: Is this one tab too many or is this how Makefile expects things? I am not fully aware of the spacing rules here.

doublek (commented):
This looks pretty neat and a pleasant Makefile PR with minimal Makefile black magic in it 😄. I had one minor nit on some spacing and one possible place where we can do some code cleanup.

Makefile.common
Brian-McM: Yikes no I don't think that has to be like that.

doublek (approved):
LGTM

Makefile.common
doublek: Discussed offline and decided to add a deprecated comment.

PR #131, opened 2020-09-10
User query params not json body for listing PRs

Brian-McM:


Makefile.common
```diff
@@ -359,14 +358,14 @@ endef
 # List pull open pull requests for a head and base. $(1) is the repo slug, $(2) is the branch head, $(3) is the branch base,
```
bcreane, line 358: comment is out of date since we're no longer using `$(1)`.

bcreane (approved):
switch basically from `:` to `=` looks fine.

Makefile.common
Brian-McM: we're still using $(1) `$(call github_call_api,GET,$(1)/pulls?$(QUERY),)` I just have it in a weird order lol

PR #132, opened 2020-09-10
Add semaphore API helpers

Brian-McM:


Makefile.common
```diff
@@ -461,6 +461,25 @@ mod-download:
 mod-tidy:
 	-$(DOCKER_RUN) $(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) go mod tidy'
 
+###############################################################################
+# Semaphore helpers
+###############################################################################
+
+# Sends a request to the semaphore API to run the request workflow. It requires setting the SEMAPHORE_API_TOKEN, SEMAPHORE_PROJECT_ID,
```
bcreane, line 468: is there a way to list semaphore_project_ids, or do a query to return the id associated with a project name string, e.g. `node-private`?

Brian-McM: Unfortunately not that I've found, it requires using the semaphore client. The project id doesn't change so I was going to set this ENV variable in every project so users don't need to figure it out.

bcreane (approved):

PR #133, opened 2020-09-11
Add Semaphore project adds for a few projects

Brian-McM:
Also allowed git remote to be configurable for git pushing.


doublek (approved):
LGTM

PR #136, opened 2020-09-15
[PICK] Use base branch as suffix to sem pin update branch, add semaphore-run-auto-pin-update-workflows target

Brian-McM:
The semaphore-run-auto-pin-update-workflows target is what a project will use to update the pins of it's dependents. A project like libcalico-go would set the SEMAPHORE_AUTO_PIN_UPDATE_PROJECT_IDS to the project ids of it's dependent projects i.e. typha then running semaphore-run-auto-pin-update-workflows would trigger the update pin pipelines each proejct defined in the ENV variable.


doublek (approved):
👍

PR #137, opened 2020-09-15
Update to go v1.15.2

caseydavenport:


Dockerfile.s390x
```diff
@@ -9,7 +9,7 @@ RUN apk --update add curl
 RUN for i in ${QEMU_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
 RUN chmod +x /usr/bin/qemu-*
 
-FROM s390x/golang:1.14.4-alpine3.11
+FROM s390x/golang:1.15.2-alpine3.11
```
fasaxc, line 12: Don't think this image exists, but there's a `1.15.2-alpine3.12`

Dockerfile.arm64
```diff
@@ -9,7 +9,7 @@ RUN apk --update add curl
 RUN for i in ${QEMU_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
 RUN chmod +x /usr/bin/qemu-*
 
-FROM arm64v8/golang:1.14.4-alpine3.11
+FROM arm64v8/golang:1.15.2-alpine3.11
```
fasaxc, line 12: Same here, looks like alpine 3.12 is the base for golang 1.15

PR #156, opened 2020-10-16
Add mockery to build image / add mockery utils

Brian-McM:
This commit adds mockery to the build image to help generate testify mocks. There are helper targets that projects can use to make sure that the mocks are regenerated if interfaces change.


bcreane (approved):
ok +`11MB`

PR #162, opened 2020-11-03
Login using docker hub, temporarily disable "run with" block, add more sem project ids

Brian-McM:
This commit does the following:
- Login into docker hub properly (was broken and not done for the test)
- Create and expose the quay secret (couldn't expose it because the orignal secret had a "+" in it and the UI complained)
- Temporarily disable only running tests when there's changes in the common makefile or in semaphore files (it's currently broken)
- Add intrusion detection, voltron, and ts queryserver sem project ids


.semaphore/semaphore.yml
```diff
@@ -39,12 +44,10 @@ blocks:
   dependencies: []
   task:
     secrets:
-    - name: quay-robot-calico+semaphoreci
```
bcreane, line 42: this is a bit of a semaphoreV2 bummer. Glad you let them know that it's broken

bcreane (approved):
/lgtm

.semaphore/semaphore.yml
Brian-McM: Ya I think they'll probably hop on it to fix it pretty quickly (I hope).

PR #172, opened 2020-12-08
Use a constant length for git abbreviations

caseydavenport:


matthewdupre (approved):
LGTM.  We'll need to coordinate the change with any other git describe calls (I think notably in hashreleases).

PR #198, opened 2020-12-30
Add docker-compress-image to compress docker image layers

Brian-McM:
This commit adds the docker-copmress-image target which exports and imports the given docker image to compress the image layers. A subset of the instructions from the uncompressed image are imported to the compressed image using the --change arguement for docker import.


doublek (approved):
LGTM

PR #224, opened 2021-03-15
[PICK] Print link to semaphore workflow when it's created

Brian-McM:
Pick of #222 (messed up the pick for that before it went to master)


doublek (approved):
LGTM

PR #225, opened 2021-04-02
Allow specifying a commit show for semaphore-run-workflow

Brian-McM:


Makefile.common
```diff
@@ -541,7 +541,7 @@ SEMAPHORE_WORKFLOW_BRANCH?=master
 semaphore-run-workflow:
 	$(eval CMD := curl -f -X POST \
 		-H "Authorization: Token $(SEMAPHORE_API_TOKEN)" \
-		-d "project_id=$(SEMAPHORE_PROJECT_ID)&reference=$(SEMAPHORE_WORKFLOW_BRANCH)&pipeline_file=.semaphore/$(SEMAPHORE_WORKFLOW_FILE)" \
+		-d "project_id=$(SEMAPHORE_PROJECT_ID)&reference=$(SEMAPHORE_WORKFLOW_BRANCH)&commit_sha=$(SEMAPHORE_COMMIT_SHA)&pipeline_file=.semaphore/$(SEMAPHORE_WORKFLOW_FILE)" \
```
doublek, line 544: Questions because of my ignorance about semv2 APIs: How does this work with both a branch reference and a commit sha?

Brian-McM: It just chooses the commit on the branch, I'm not entirely sure how that's different from just using the sha but the ref is required.

Brian-McM: The might be different rules for running a pipeline for the same commit but on a differet branch?

doublek: And if we don't specify a commit it just runs the HEAD of that branch?

Brian-McM: Yes exactly, I tested that.

doublek (approved):
LGTM

PR #226, opened 2021-04-02
Allow specifying a commit show for semaphore-run-workflow

Brian-McM:


doublek:
Retroactive LGTM based on #225

PR #236, opened 2021-04-13
Don't fail if pin update PR already exists, just exit with a message

Brian-McM:
Pick of https://github.com/projectcalico/go-build/pull/235


doublek (approved):
LGTM

PR #239, opened 2021-04-22
Add replace / update apiserver pins

Brian-McM:


Makefile.common
```diff
@@ -299,6 +301,12 @@ FELIX_REPO?=github.com/projectcalico/felix
 CNI_BRANCH?=$(PIN_BRANCH)
 CNI_REPO?=github.com/projectcalico/cni-plugin
 
+update-typha-pin:
```
bcreane, line 304: should this replace the same target name on line 310?

Brian-McM: woops, thanks man.

bcreane (approved):

PR #242, opened 2021-04-26
Fix missing variant for the manifest-tool template

aalaesar:
This allow the manifest-tool to find the proper arm image for any component


caseydavenport (approved):
Seems to work for me - thanks!

PR #244, opened 2021-04-28
Don't use --long for GIT_VERSION

caseydavenport:


Makefile.common
```diff
@@ -741,7 +741,7 @@ endif
 # and BRANCH_NAME env variables to figure out what to tag and where to push it to.
 cd-common: require-confirm require-branch-name
 	$(MAKE) tag-images-all push-all push-manifests push-non-manifests IMAGETAG=$(BRANCH_NAME) EXCLUDEARCH="$(EXCLUDEARCH)"
-	$(MAKE) tag-images-all push-all push-manifests push-non-manifests IMAGETAG=$(GIT_VERSION) EXCLUDEARCH="$(EXCLUDEARCH)"
+	$(MAKE) tag-images-all push-all push-manifests push-non-manifests IMAGETAG=$(shell git describe --tags --dirty --long --always --abbrev=12) EXCLUDEARCH="$(EXCLUDEARCH)"
```
doublek, line 744: Maybe needs to be moved up to a variable with a comment explaining intended use?

doublek: Just in case we end up using this somewhere else. Hashrelease maybe? But perhaps no other place.

caseydavenport: Yeah, I considered that, but after thinking about it I felt like this was really the best place for it - I think putting it in a var would mean that other places might start using it when they don't intend to, but that's a bit of FUD.

Right now I think this is the only place that needs it - the hashreleases do expect this format, but the don't import Makefile.common.

mgleung (approved):
LGTM

doublek (approved):
LGTM

PR #247, opened 2021-04-29
Add release target helpers

Brian-McM:
This commit adds the targets that the repos will need for cutting releases. It includes the following:
- Targets to create new release branches and bump the dev tag for the next release on master
- Create release tags and annotate the current commit with that tag
- Retage docker images from their dev tag to their release tag and move them to the release repositories

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -94,6 +94,9 @@ ifeq ($(RELEASE),true)
 PUSH_IMAGES+=$(RELEASE_IMAGES)
 endif
 
+# Convenience function to get the first dev image repo in the list.
+DEV_REGISTRY ?= $(firstword $(DEV_REGISTRIES))
+
 # remove from the list to push to manifest any registries that do not support multi-arch
 EXCLUDE_MANIFEST_REGISTRIES ?= quay.io/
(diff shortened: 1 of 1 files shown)


Makefile.common
```diff
@@ -739,10 +757,167 @@ endif
 
 # cd-common tags and pushes images with the branch name and git version. This target uses PUSH_IMAGES, BUILD_IMAGE,
 # and BRANCH_NAME env variables to figure out what to tag and where to push it to.
-cd-common: require-confirm require-branch-name
+cd-common: var-require-one-of-CONFIRM-DRYRUN var-require-all-BRANCH_NAME
 	$(MAKE) tag-images-all push-all push-manifests push-non-manifests IMAGETAG=$(BRANCH_NAME) EXCLUDEARCH="$(EXCLUDEARCH)"
 	$(MAKE) tag-images-all push-all push-manifests push-non-manifests IMAGETAG=$(shell git describe --tags --dirty --long --always --abbrev=12) EXCLUDEARCH="$(EXCLUDEARCH)"
 
+###############################################################################
+# Release targets and helpers
```
doublek, line 765: I think we should add some comments about what is a 
- dev tag
- release tag
- dev image
- release image

doublek (approved):
Just one request for comment. Otherwise LGTM 🎉

PR #252, opened 2021-05-17
Support multi build images, multiple registries, and multiple arches in ci / cd / release targets

Brian-McM:


Makefile.common
```diff
@@ -351,14 +351,37 @@ endif
 git-commit:
 	git diff --quiet HEAD || git commit -m "Semaphore Automatic Update" go.mod go.sum $(EXTRA_FILES_TO_COMMIT)
 
+###############################################################################
+# External resource affecting macros
+# The following macros affect resources outside of the local environment that
+# they're run in, i.e. pushing to docker or github. If DRYRUN is defined, then
```
doublek, line 357: I think  dry run behavior be the default. I think I missed pointing this out in the previous PR as well. Elsewhere we have something similar in spirit with the requirement to specify a `CONFIRM` variable to actually push images.

Makefile.common
```diff
@@ -98,9 +98,10 @@ endif
 DEV_REGISTRY ?= $(firstword $(DEV_REGISTRIES))
 
 # remove from the list to push to manifest any registries that do not support multi-arch
-EXCLUDE_MANIFEST_REGISTRIES ?= quay.io/
-PUSH_MANIFEST_IMAGES=$(PUSH_IMAGES:$(EXCLUDE_MANIFEST_REGISTRIES)%=)
-PUSH_NONMANIFEST_IMAGES=$(filter-out $(PUSH_MANIFEST_IMAGES),$(PUSH_IMAGES))
+NONMANIFEST_REGISTRIES      ?= quay.io
```
doublek, line 101: I though quay now supported pushing multiarch/os manifests now. If so do we still want to keep this? Perhaps for some compatibility reasons?

doublek (commented):
Couple of comments around quay and dry run default behavior.

Makefile.common
Brian-McM: I'm just trying to keep things the same as they were before, so I don't want to change this in this PR.

Makefile.common
Brian-McM: That's a good point, I'll update that.

Makefile.common
doublek: Got it.

doublek (approved):
LGTM

doublek (approved):
LGTM

PR #256, opened 2021-05-21
Fix using REGISTRY instead of registry for PUSH_MANIFEST_IMAGES

Brian-McM:


Makefile.common
```diff
@@ -107,7 +107,7 @@ DEV_REGISTRY ?= $(firstword $(DEV_REGISTRIES))
 NONMANIFEST_REGISTRIES      ?= quay.io
 MANIFEST_REGISTRIES         ?= $(DEV_REGISTRIES:$(NONMANIFEST_REGISTRIES)%=)
 
-PUSH_MANIFEST_IMAGES := $(foreach registry,$(MANIFEST_REGISTRIES),$(foreach image,$(BUILD_IMAGES),$(call filter-registry,$(REGISTRY))$(image)))
+PUSH_MANIFEST_IMAGES := $(foreach registry,$(MANIFEST_REGISTRIES),$(foreach image,$(BUILD_IMAGES),$(call filter-registry,$(registry))$(image)))
```
stevegaossou, line 110: It looks like every other reference to that variable is in all-caps though.

[Here](https://github.com/projectcalico/go-build/blob/e53d5a2516a25dfbdc87f25aa512f91e11ef1212/Makefile.common#L732-L739) and [here](https://github.com/projectcalico/go-build/blob/e53d5a2516a25dfbdc87f25aa512f91e11ef1212/Makefile.common#L759-L766).

Brian-McM: It's a different variable, it's a for loop so this is referencing registry in `$(foreach registry,$(MANIFEST_REGISTRIES)`

stevegaossou (approved):
LGTM

mgleung (approved):
LGTM

PR #298, opened 2021-07-20
Added check-module-path static check target

song-jiang:
This PR added two targets to make sure that correct API module is imported.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -564,6 +564,38 @@ install-git-hooks:
 foss-checks:
 	$(DOCKER_RUN) -e FOSSA_API_KEY=$(FOSSA_API_KEY) $(CALICO_BUILD) /usr/local/bin/fossa
 
+.PHONY: check-module-path-tigera-api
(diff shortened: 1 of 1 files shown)


Makefile.common
```diff
@@ -564,6 +564,34 @@ install-git-hooks:
 foss-checks:
 	$(DOCKER_RUN) -e FOSSA_API_KEY=$(FOSSA_API_KEY) $(CALICO_BUILD) /usr/local/bin/fossa
 
+IMPORT_TIGERA_API = $(shell go list -m github.com/tigera/api > /dev/null 2>&1 && echo yes || echo no)
```
Brian-McM, line 567: This should probably be implemented as a macro and called by the targets. This implementation means that this is overhead run for everything the imports this Makefile (which is more than just go build projects, contrary to the name of this project).

Brian-McM (changes requested):

Brian-McM (approved):

PR #314, opened 2021-08-19
Fix: debian stable release change.

frozenprocess:
Debian changed stable release to bullseye.
@fasaxc

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -25,12 +25,12 @@ ARG MANIFEST_TOOL_VERSION=v1.0.2
 # Install grep, sed, zip, and jq for use in some Makefiles (e.g. pulling versions out of glide.yaml)
 # Install gcc for cgo.
 # Install clang, libbpf and newer kernel headers for building BPF binaries.
-RUN echo 'APT::Default-Release "stable";' > /etc/apt/apt.conf.d/99defaultrelease && \
+RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease && \
     echo 'deb     http://ftp.am.debian.org/debian/    buster-backports main contrib non-free' > /etc/apt/sources.list.d/buster-backports.list && \
     apt-get -y update &&  \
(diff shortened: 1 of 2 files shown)


Dockerfile.amd64
```diff
@@ -25,12 +25,12 @@ ARG MANIFEST_TOOL_VERSION=v1.0.2
...
 # Install gcc for cgo.
 # Install clang, libbpf and newer kernel headers for building BPF binaries.
-RUN echo 'APT::Default-Release "stable";' > /etc/apt/apt.conf.d/99defaultrelease && \
+RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease && \
     echo 'deb     http://ftp.am.debian.org/debian/    buster-backports main contrib non-free' > /etc/apt/sources.list.d/buster-backports.list && \
     apt-get -y update &&  \
     apt-get -y upgrade && \
     apt-get install --no-install-recommends -y -t buster-backports \
-        libbpf-dev linux-headers-amd64  && \
+        libbpf-dev linux-headers-5.10.0-0.bpo.8-amd64  && \
```
fasaxc, line 33: Why is this pin needed?  What version were we on before?

frozenprocess: Atm `linux-headers-amd64` from `buster-backports` is using a different kernel header resulting in the following error.
```
The following packages have unmet dependencies:
 linux-headers-amd64 : Depends: linux-headers-5.10.0-0.bpo.8-amd64 (= 5.10.46-2~bpo10+1) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```
Since I was not sure what could be the implication of updating all the related stuff tried to take the safe route by using this pinned version from [last successful build](https://tigera.semaphoreci.com/jobs/080a1683-5eb7-485d-9b68-e9a0e263e6e6#L430).

fasaxc: Sounds reasonable

fasaxc (approved):

PR #315, opened 2021-08-19
Include code generation tools in go-build image

caseydavenport:


Dockerfile.amd64
```diff
@@ -87,8 +87,16 @@ RUN go get github.com/Masterminds/glide && \
   go get -u golang.org/x/vgo && \
   go get -u github.com/jstemmer/go-junit-report && \
   go get -u golang.org/x/tools/cmd/stringer && \
+  go get -u k8s.io/code-generator/cmd/openapi-gen && \
+  go get -u k8s.io/code-generator/cmd/deepcopy-gen && \
   rm -rf /go/src/* /root/.cache
 
+# Used for generating CRD files.
```
caseydavenport, line 94: We only have an amd64 build of this, but I think it's OK because it's just used as part of the generation and not needed for cross-compiling.

Probably same goes for the other two generation tools, but I included them anyway because it's easy.

fasaxc (approved):

PR #327, opened 2021-09-07
[v0.55] Auto pick #326: Add skimble and DPI sem IDs

Brian-McM:
Cherry pick of #326 on v0.55.

#326: Add skimble and DPI sem IDs


doublek (approved):
LGTM

PR #338, opened 2021-09-10
Support only cutting an image and allow for image prefixes

Brian-McM:
This commit does the following:
1. Supports the IMAGETAG_PREFIX when cutting the images
2. Allows a user to only cut an image (not push git tags) when cutting the release using the IMAGE_ONLY env variable
3. Support verifying the release tag is an expected tag with the optional EXPECTED_RELEASE_TAG env variable

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -954,15 +954,26 @@ cut-release-wrapped: var-require-one-of-CONFIRM-DRYRUN
 	$(eval DEV_TAG = $(call git-dev-tag))
 	$(eval RELEASE_TAG = $(call git-release-tag-from-dev-tag))
 	$(eval RELEASE_BRANCH = $(call release-branch-for-tag,$(DEV_TAG)))
+ifdef EXPECTED_RELEASE_TAG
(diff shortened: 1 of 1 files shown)


Makefile.common
```diff
@@ -954,15 +954,26 @@ cut-release-wrapped: var-require-one-of-CONFIRM-DRYRUN
 	$(eval DEV_TAG = $(call git-dev-tag))
 	$(eval RELEASE_TAG = $(call git-release-tag-from-dev-tag))
 	$(eval RELEASE_BRANCH = $(call release-branch-for-tag,$(DEV_TAG)))
+ifdef EXPECTED_RELEASE_TAG
```
bcreane, line 957: is `EXPECTED_RELEASE_TAG` documented anywhere?

bcreane (approved):
we can document the new vars if they're not already described elsewhere in a subsequent PR.

Makefile.common
Brian-McM: No it's not documented

PR #340, opened 2021-09-17
Verify release before triggering cut in semaphore, trigger test-cut-release if CONFIRM is not specified

Brian-McM:

PR #341, opened 2021-09-17
[v0.55] Auto pick #340: Verify release before triggering cut in semaphore, trigger

Brian-McM:
Cherry pick of #340 on v0.55.

#340: Verify release before triggering cut in semaphore, trigger


doublek (approved):
LGTM

PR #344, opened 2021-09-23
Update to golang v1.16, k8s v1.22

caseydavenport:


Dockerfile.amd64
```diff
@@ -15,14 +15,14 @@ ARG MANIFEST_TOOL_VERSION=v1.0.2
 
 # Install su-exec for use in the entrypoint.sh (so processes run as the right user)
 # Install bash for the entry script (and because it's generally useful)
-# Install curl to download glide
+# Install curl
```
bcreane, line 18: I'm glad removing `glide` is an option - I was hitting some `go get` conflicts because it's so old and un-loved.

Dockerfile.armv7
```diff
@@ -43,21 +43,9 @@ ENV CGO_ENABLED=0
...
-# Install dep
-RUN go get github.com/golang/dep/cmd/dep
-
 # Install ginkgo CLI tool for running tests
 RUN go get github.com/onsi/ginkgo/ginkgo
 
-# Install linting tools.
-RUN go get -u gopkg.in/alecthomas/gometalinter.v2
-RUN ln -s `which gometalinter.v2` /usr/local/bin/gometalinter
-RUN gometalinter --install
```
bcreane, line 59: is linter now part of the base image?

bcreane (approved):
Looks good - nice to see all those red blocks of removed. Couple of questions, but approved anyway.

Dockerfile.armv7
caseydavenport: I'm not really sure - I removed it because it looked like we didn't pull it into the amd64 image, and it was failing to build with go 1.16

caseydavenport:
I'm gonna merge and try this and we can always add gometalinter back if it's actually used.

PR #353, opened 2021-11-29
Include ModSecurity dependencies and build for Dikastes [WAF]

StevenTigera:


scripts/modsec.sh
```diff
@@ -0,0 +1,11 @@
...
+set -e
+cd /tmp
+git clone https://github.com/SpiderLabs/ModSecurity.git
+cd ModSecurity
+git checkout ${MODSEC_VERSION:-v3.0.6}
+git submodule update --init
+./build.sh
+./configure
+make
+make install
```
fasaxc, line 11: If the files in the ModSecurity dir are not needed after install then you should delete the directory.  Otherwise all the intermediate object files will be stored in the docker image

scripts/modsec.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env sh
+set -e
+cd /tmp
+git clone https://github.com/SpiderLabs/ModSecurity.git
+cd ModSecurity
+git checkout ${MODSEC_VERSION:-v3.0.6}
```
fasaxc, line 6: +1 for pinning the version.

fasaxc (changes requested):
Couple of minor points from me.  Does modsec depend on openssl?  ( see the libcurl4-openssl dependency)  If so, it might have an impact on @caseydavenport 's FIPS work.

Dockerfile.amd64
```diff
@@ -60,6 +60,15 @@ RUN chmod +x /usr/local/bin/mockery
 RUN echo 'Host *' >> /etc/ssh/ssh_config \
   && echo '    StrictHostKeyChecking no' >> /etc/ssh/ssh_config
 
+# ModSecurity dependencies.
+RUN apt-get -y update &&  \
+    apt-get -y upgrade && \
+    apt-get install --no-install-recommends -y \
+        apt-utils autoconf automake build-essential \
```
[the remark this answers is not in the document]
StevenTigera: Resolved in commit 194a7ba60538e0a3303f1a31386912c4eecd4db5

scripts/modsec.sh
StevenTigera: Resolved in commit 194a7ba60538e0a3303f1a31386912c4eecd4db5

fasaxc (approved):
LGTM.

PR #356, opened 2021-12-01
Update goboring version to 1.16.10 to squish CVE's

Brian-McM:


matthewdupre (approved):
Fine assuming that there are no other associated files in the image that need to be updated

PR #368, opened 2022-01-28
feat(Makefile.common): add update_replace_submodule_pin

freecaykes:
update_replace_submodule_pin is used similarly as update_replace_pin without retrieving the git version of the replacing repo. It would take parameted $(3) as the version used by go mod edit. Used for updating pins for repos with a path. as git cannot resolve a version given a repo path deeper than the top repo folder.


Makefile.common
```diff
@@ -310,6 +310,22 @@ define update_replace_pin
 		fi'
 endef
 
+# update_replace_pin_with_path is used similarly as update_replace_pin without retrieving the git version of the 
+# replacing repo. It would take parameted $(3) as the version used by go mod edit. Used for updating pins for
+# repos with a path. as git cannot resolve a version given a repo path deeper than the top repo folder.
```
doublek, line 315: couple of nit on the comment:
1. Rather than "git version", perhaps "git commit SHA" is better/specific? 
2. Typo parameted -> parameters
3.  Rather than "Repos with a path", it should probably say "modules for directories within a repository" perhaps?

doublek: Would also like to see the parameters defined here as well so that I don't have to look up the ones in update_replace_pin.

doublek (approved):
LGTM

PR #383, opened 2022-03-25
Build images on tag triggers as well

doublek:
- [ ] Update README/Add RELEASING
- [ ] Remove branch based image pushes
- [ ] Check tagged image doesn't exist before pushing
- [ ] Update downstream tests

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -59,9 +59,9 @@ blocks:
       - env_var: ARCH
         values: ["amd64", "arm64","armv7", "ppc64le"]
 
-- name: "Push manifest"
+- name: "Push manifest (branch)"
   skip:
-    # Only run on branches, not PRs.
+    # Only run on branches, not PRs or tags.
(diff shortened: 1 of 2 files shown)


.semaphore/semaphore.yml
```diff
@@ -59,9 +59,9 @@ blocks:
       - env_var: ARCH
         values: ["amd64", "arm64","armv7", "ppc64le"]
 
-- name: "Push manifest"
+- name: "Push manifest (branch)"
```
caseydavenport, line 62: We can probably get rid of this altogether on master now - new releases will just be done via tags instead. WDYT?

doublek: That is a reasonable thing to do. The reason for leaving it behind (although a tad lazy and I will clean it up) was  thought it would provide someone to use (at their own risk) a branch image and don't have to keep bumping the image. Given that it would take away from the whole purpose of tags, I am okay to remove it. Removes the copy/paste cruft that I did here as well.

PR #388, opened 2022-04-21
use ginkgo v2

tomastigera:


caseydavenport (approved):
@tomastigera LGTM - the only thought might be to switch it around so that ginkgo2 is the default, and ginkgo1 is opt-in.

That's a bit more work up-front, but it would be nice to stop adding new ginkgo v1 runs for new test suites. 

Not too fussed about it, so long as we don't sit on both versions for too long.

PR #393, opened 2022-05-17
Fix misplaced quote at the end of release-retag-dev-image-in-registry-%

Brian-McM:


Brian-McM:
Force merged because master is already failing with the same issue and it was a sporadic one that popped up.

PR #397, opened 2022-05-20
Add MOCKERY_IN_PACKAGE option (default to true to create mocks in package, otherwise go to mocks dir)

Brian-McM:


caseydavenport (approved):
Approved, but you need to add comments for MOCKERY_IN_PACKAGE explaining what its purpose is and why you might want to set it one way or another. 

As I understand it, the use case is basically "one repo has done something differently" so we should also agree on what we intend to standardize on, and when / why a repo might want to break from that standard.

PR #398, opened 2022-05-20
go-1.18

tomastigera:


fasaxc (approved):

Dockerfile.s390x
```diff
@@ -44,27 +44,23 @@ ENV CGO_ENABLED=0
 RUN go install -v std
 
 # Install ginkgo CLI tool for running tests
-RUN go get github.com/onsi/ginkgo/ginkgo
-
-# Install linting tools.
-RUN wget -O - -q https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.20.0
-RUN golangci-lint --version
```
Brian-McM, line 51: Are we not installing any linter now? I think that will breaker the static checks...

PR #400, opened 2022-06-30
Consider tigera packages as local packages

gantony:
We currently treat packages starting with `github.com/projectcalico/` as local ones, which formats their import in a different section. This change adds the packages starting with `github.com/tigera/`to the list of local ones so that they're listed below the ones from `github.com/projectcalico/`.

Default formatting, without any `-local` flag:
```
import (
	"flag"

	"github.com/projectcalico/calico/libcalico-go/lib/logutils"
	"github.com/sirupsen/logrus"
	updater "github.com/tigera/runtime-security/pkg/sasha-updater"
	verify "github.com/tigera/runtime-security/pkg/verify"
)
```

Current solution, using `-local github.com/projectcalico/`:
```
import (
	"flag"

	"github.com/sirupsen/logrus"
	updater "github.com/tigera/runtime-security/pkg/sasha-updater"
	verify "github.com/tigera/runtime-security/pkg/verify"

	"github.com/projectcalico/calico/libcalico-go/lib/logutils"
)
```

Proposed change, using `-local github.com/projectcalico/,github.com/tigera/`:
```
import (
	"flag"

	"github.com/sirupsen/logrus"

	"github.com/projectcalico/calico/libcalico-go/lib/logutils"
	updater "github.com/tigera/runtime-security/pkg/sasha-updater"
	verify "github.com/tigera/runtime-security/pkg/verify"
)
```

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -573,11 +573,11 @@ golangci-lint: $(GENERATED_FILES)
 
 .PHONY: go-fmt goimports fix
 fix go-fmt goimports:
-	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'find . -iname "*.go" ! -wholename "./vendor/*" | xargs goimports -w -local github.com/projectcalico/'
+	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'find . -iname "*.go" ! -wholename "./vendor/*" | xargs goimports -w -local github.com/projectcalico/,github.com/tigera/'
 
 check-fmt:
 	@echo "Checking code formatting.  Any listed files don't match goimports:"
-	$(DOCKER_RUN) $(CALICO_BUILD) bash -c 'exec 5>&1; ! [[ `find . -iname "*.go" ! -wholename "./vendor/*" | xargs goimports -l -local github.com/projectcalico/ | tee >(cat >&5)` ]]'
+	$(DOCKER_RUN) $(CALICO_BUILD) bash -c 'exec 5>&1; ! [[ `find . -iname "*.go" ! -wholename "./vendor/*" | xargs goimports -l -local github.com/projectcalico/,github.com/tigera/ | tee >(cat >&5)` ]]'
 
 .PHONY: pre-commit
 pre-commit:


caseydavenport:
I suspect we probably want to make this configurable rather than hard-code to this. What is local and what is not local depends on the context of the repository that the target is being run against, so it would make sense for each repo to make that call for itself.

Maybe we can auto-detect from the git config? Or just have other repos configure this with an env var?

neiljerram:
@caseydavenport I think you're suggesting that there should be a grouping for "all of this repo's own packages"; is that right?  So if the repo is github.com/x/y, that maps to `-local github.com/x/y`.

Another option would be to remove the `-local` entirely, which means our packages get mixed in with any others.  Any thoughts on that one?

gantony:
> I suspect we probably want to make this configurable rather than hard-code to this.

This would work and would be easy to do. It would be an opt-in for the repo using it. We could also make it configurable, with default value of `-local github.com/projectcalico/,github.com/tigera/`, and people can change it if it does not suit.

> @caseydavenport I think you're suggesting that there should be a grouping for "all of this repo's own packages"; is that right? 

I thought about that initially, and I could use `$(PACKAGE_NAME)` for that. But then, why ignore other `tigera` packages?

I like the proposed solution because it's simple and works well in most scenarios (although I probably miss some context here). 
My mental model is that `tigera` repos are private and `calico` ones are public. So generally speaking the `calico` repos won't be impacted by the change (as they don't import `tigera` repos), except for a bit of `noise` in the `Makefile.common`. 

The only users I could see being impacted would be some OS calico user who may want to include other packages as local. In that case the suggestion above to make it configurable, with default value of `-local github.com/projectcalico/,github.com/tigera/` would work. But that seems a bit complex for little value. Again, I'm probably missing some context here. 

For example, is `Makefile.common` typically used to build repos that are neither in `github.com/projectcalico` or `github.com/tigera`? This, I understand, would be the main motivation for making the setting configurable.

caseydavenport:
> My mental model is that tigera repos are private and calico ones are public

This isn't strictly true - we do have some public tigera/ repositories and private projectcalico/ repositories (although they are definitely the exception rather than the rule). 

> But then, why ignore other tigera packages?

I suppose it depends on what our goal with this change is :grin: 

- Group all of the packages from the local repository together, and remote dependencies separately, so it's easier to see what is a local import?
- Group all of the packages that "Tigera" as a company maintains together, and the rest separately?
- Something else?

I will say, I also don't really know the history behind setting this option the way it is today...

> Another option would be to remove the -local entirely, which means our packages get mixed in with any others. Any thoughts on that one?

I don't really have a problem with this other than switching from one approach to the other unnecessarily causes a large number of lines of code to change unnecessarily.

I guess I'm mostly just providing a bit of pressure, but don't really understand the motivation or have too much skin in the game otherwise :laughing: 

Would be nice to not needlessly churn our imports, but adding github.com/tigera/ to -local shouldn't do that for projectcalico/calico at least.

caseydavenport:
> For example, is Makefile.common typically used to build repos that are neither in github.com/projectcalico or github.com/tigera? This, I understand, would be the main motivation for making the setting configurable.

Nope, I think its use is entirely within those two organizations, so not a big deal really. I think it comes back to me not being sure what the behavior that we want is. My instinct is "group local imported packages together and anything else separately", which means even `github.com/projectcalico/` imports from a different repo shouldn't be grouped with imports from the same repo. But this is all effectively cosmetics at that point so.... :shrug:

caseydavenport:
^ If you're OK with all of that and don't see any problems, then you can definitely ignore me, I don't intend to block this if you think it will be valuable for what you're working on.

PR #402, opened 2022-08-12
Update Golang 1.18.5 for amd64

ry4nz:
Go boring crypto bits location is changed according to
https://go.googlesource.com/go/+/refs/heads/dev.boringcrypto/misc/boring/README.md

This PR updates the golang version for amd64

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -1,11 +1,9 @@
 FROM calico/bpftool:v5.3-amd64 as bpftool
 
-FROM goboring/golang:1.16.7b7
+FROM us-docker.pkg.dev/google.com/api-project-999119582588/go-boringcrypto/golang:1.18.5b7
 MAINTAINER Shaun Crampton <shaun@projectcalico.org>
 
-# Override the go boring version in the image with the latest release. There is currently no 1.17 container image,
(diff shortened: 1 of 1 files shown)


Dockerfile.amd64
```diff
@@ -1,11 +1,9 @@
 FROM calico/bpftool:v5.3-amd64 as bpftool
 
-FROM goboring/golang:1.16.7b7
+FROM us-docker.pkg.dev/google.com/api-project-999119582588/go-boringcrypto/golang:1.18.5b7
```
caseydavenport, line 3: What is the source for this image?

For previous updates, adjusting the GO_BORING_VERSION below and sticking with the official `goboring/` images was enough. Do we need this part of the change?

ry4nz: Based on https://go.googlesource.com/go/+/refs/heads/dev.boringcrypto/misc/boring/README.md the images are no longer hosted at docker hub `docker.io/goboring/golang`.     They are hosted at the above location.

ry4nz: Updating the base image from 1.16.7 to 1.18.5 to fix CVEs that comes with the 1.16.7 image.

caseydavenport: Great, thanks for that link.

PR #407, opened 2022-10-05
linux-headers package update.

frozenprocess:
This PR addresses the `buster-backports` linux-headers package update.


caseydavenport (approved):

frozenprocess:
@caseydavenport seems like `buster-backports` repo is not serving two kernel packages anymore, which prompted us to pin the package in the first place.

buster repo: https://packages.debian.org/search?suite=buster-backports&arch=any&searchon=names&keywords=linux-headers-5.10 

I would say it is safe to use the more generic package name, WDYT?

https://github.com/projectcalico/go-build/pull/314#discussion_r691911358 <- original package problem

caseydavenport:
Yep, that sounds reasonable to me. @frozenprocess

PR #408, opened 2022-10-27
add swaggo to dockerfiles

WilliamTigera:
Add swagger and update ginkgo v2

### DIFF
--- a/Dockerfile.armv7
+++ b/Dockerfile.armv7
@@ -70,9 +70,12 @@ RUN  go install k8s.io/code-generator/cmd/deepcopy-gen@master
 
 # Install ginkgo v2 as ginkgo2 and keep ginkgo v1 as ginkgo
 RUN mv /go/bin/ginkgo /go/bin/ginkgo1 && \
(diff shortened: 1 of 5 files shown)


Dockerfile.armv7
```diff
@@ -70,10 +70,13 @@ RUN  go install k8s.io/code-generator/cmd/deepcopy-gen@master
 
 # Install ginkgo v2 as ginkgo2 and keep ginkgo v1 as ginkgo
 RUN mv /go/bin/ginkgo /go/bin/ginkgo1 && \
-  go install github.com/onsi/ginkgo/v2/ginkgo@v2.1.3 && \
+  go install github.com/onsi/ginkgo/v2/ginkgo@v2.3.0 && \
   mv /go/bin/ginkgo /go/bin/ginkgo2 && \
   mv /go/bin/ginkgo1 /go/bin/ginkgo
 
+# Install Swaggo
+RUN go install github.com/swaggo/swag/cmd/swag@latest
```
Brian-McM, line 78: I don't think we should install latest, we should always have a specific version in mind. The reason is that latest could update to have breaking changes.

Brian-McM (approved):

PR #419, opened 2023-02-14
Auto-update golang to latest 1.19 

Behnam-Shobiri:
This PR will always use the (latest-1) minor version's latest patch. For example, now it would use the golang 1.19.6.
The only downside is that when golang 1.21 is out we have to fix it on the latest 1.19 and create a new branch then. I think this is much better than creating a branch with every patch release.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -3,7 +3,7 @@ FROM calico/bpftool:v5.3-amd64 as bpftool
 FROM us-docker.pkg.dev/google.com/api-project-999119582588/go-boringcrypto/golang:1.18.10b7
 MAINTAINER Shaun Crampton <shaun@projectcalico.org>
 
-ARG GO_VERSION=1.19.6
+ARG GO_VERSION
 ARG QEMU_VERSION=4.2.0-6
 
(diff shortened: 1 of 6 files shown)


Dockerfile.amd64
```diff
@@ -1,9 +1,9 @@
 FROM calico/bpftool:v5.3-amd64 as bpftool
 
-FROM us-docker.pkg.dev/google.com/api-project-999119582588/go-boringcrypto/golang:1.18.9b7
+FROM us-docker.pkg.dev/google.com/api-project-999119582588/go-boringcrypto/golang:1.18.10b7
```
caseydavenport, line 3: Should this not be 1.19?

rene-dekker: I think that all that matters here is that the base image has b7. There is no golang:1.19 base image. We could create one though, as the dockerfile is in https://pkg.go.dev/ somewhere.

rene-dekker (approved):
lgtm

hjiawei:
Replaced by https://github.com/projectcalico/go-build/pull/423.

PR #430, opened 2023-04-01
Cleanup GOROOT copies and dependencies

hjiawei:
This changeset reduces go-build image size by not making another `GOROOT` copy for `CGO` and cleaning up go mod cache. For amd64, this saves ~1.55 GB uncompressed. The uncompressed amd64 image size is down to 3.9 GB (from 5.44 GB in v0.83).

Here is the list of detailed changes:
* Don't copy `GOROOT` to `GOCGO` and `GOCGO` is removed. `boringcrypto` is integrated into the main branch from Go 1.19 so we probably don't need the second copy.
  - save ~950 MB.
* Don't install Go standard libraries with `CGO_ENABLED=0` flag. With recent FIPS work, Calico OSS and Enterprise build process will selectively enable or disable CGO per component, installing standard libraries with CGO disabled isn't needed. Without this step in go-build, we can still get statically linked binaries.
* Clean up go mod cache.
  - save ~650 MB.
* Remove `mercurial`, no errors fetching go dependencies with `git` alone.
* `bash` is installed by default in debian base image so don't list it in `apt-get`.
* Remove `fossa`.
* Simplified LLVM and ModSecurity installation steps.
* Upgrade k8s utilities to the latest v1.25 release train.


Dockerfile.amd64
```diff
@@ -34,20 +33,19 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
     apt-get install --no-install-recommends -y -t buster-backports \
         libbpf-dev linux-headers-amd64  && \
     apt-get install --no-install-recommends -y \
-        curl bash git openssh-client mercurial make wget util-linux file grep sed jq zip \
-        lsb-release software-properties-common binutils file iproute2 \
```
hjiawei, line 38: `file` is listed in the previous line so removed.

Dockerfile.amd64
```diff
@@ -34,20 +33,19 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
...
+        curl git openssh-client make wget util-linux file grep sed jq zip \
+        lsb-release software-properties-common binutils iproute2 \
         ca-certificates gcc mingw-w64 libc-dev bsdmainutils strace libpcap-dev \
         apt-utils autoconf automake build-essential \
         libcurl4-openssl-dev libgeoip-dev liblmdb-dev \
         libpcre++-dev libtool libxml2-dev libyajl-dev \
-        pkgconf zlib1g-dev && \
-    rm -rf /var/lib/apt/lists/*
```
hjiawei, line 44: `apt` cleanups are moved after LLVM install.

hjiawei:
For amd64, compressed image size down to 1.4 GB from 1.8 GB in v0.83.

Dockerfile.amd64
```diff
@@ -151,8 +132,11 @@ RUN wget https://github.com/google/go-containerregistry/releases/download/v0.4.1
 COPY --from=bpftool /bpftool /usr/bin
 
 # Build ModSecurity for Dikastes.
-COPY scripts/modsec.sh /usr/local/bin/scripts/modsec.sh
-RUN /usr/local/bin/scripts/modsec.sh
+ENV MODSEC_VERSION=v3.0.8
+RUN git clone -b ${MODSEC_VERSION} --depth 1 --recurse-submodules --shallow-submodules https://github.com/SpiderLabs/ModSecurity.git /build && \
```
hjiawei, line 136: Simplified from `scripts/modsec.sh`.

hjiawei:
Closing for now to do more multiarch build tests.

hjiawei:
Replaced by https://github.com/projectcalico/go-build/pull/433.

PR #432, opened 2023-04-03
Allow release version override

danudey:
Small update to the Makefile to allow overriding the release value check if necessary.


caseydavenport:
@danudey this one is over a year old - do we still need it?

PR #433, opened 2023-04-03
Cleanup GOROOT copies and dependencies

hjiawei:
This changeset reduces go-build image size by not making another `GOROOT` copy for `CGO` and cleaning up go install/mod cache. For amd64, this saves ~1.9 GB uncompressed. The uncompressed amd64 image size is down to 3.51 GB (from 5.44 GB in v0.83). Best efforts to match other arches with amd64/arm64.

Here is the list of detailed changes:
* Don't copy `GOROOT` to `GOCGO` and `GOCGO` is removed. `boringcrypto` is integrated into the main branch from Go 1.19 so we probably don't need the second copy.
  - save ~950 MB.
* Don't install Go standard libraries with `CGO_ENABLED=0` flag. With recent FIPS work, Calico OSS and Enterprise build process will selectively enable or disable CGO per component, installing standard libraries with CGO disabled isn't needed. Without this step in go-build, we can still get statically linked binaries.
  - save ~400 MB.
* Clean up go mod cache.
  - save ~650 MB.
* Remove `mercurial`, no errors fetching go dependencies with `git` alone.
* `bash` is installed by default in debian base image so don't list it in `apt-get`.
* Remove `fossa`.
* Simplified LLVM and ModSecurity installation steps.
* Upgrade k8s utilities to the latest v1.25 release train.
* Upgrade qemu to v7.2.0 (stuck locally when building armv7 using qemu v4.2.0; upgrading qemu seems to fix the issue.).


Dockerfile.amd64
```diff
@@ -34,125 +34,91 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
...
-    rm -rf /var/lib/apt/lists/*
+        pkgconf zlib1g-dev
 
-RUN rm /etc/apt/sources.list.d/buster-backports.list
+RUN curl -sfL https://apt.llvm.org/llvm.sh | bash -s -- ${LLVM_VERSION} && \
+    apt-get install clang-${LLVM_VERSION}
 
-RUN wget https://apt.llvm.org/llvm.sh
-RUN bash ./llvm.sh 12
-RUN apt install clang-12
```
hjiawei, line 50: Reduced image layers above.

Dockerfile.amd64
```diff
@@ -34,125 +34,91 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
...
 # su-exec is used by the entrypoint script to execute the user's command with the right UID/GID.
 # (sudo doesn't work easily in a container.)  The version was current master at the time of writing.
-ARG SU_EXEC_VER=212b75144bbc06722fbd7661f651390dc47a43d1
 RUN  set -ex; \
-     curl -o /sbin/su-exec.c https://raw.githubusercontent.com/ncopa/su-exec/${SU_EXEC_VER}/su-exec.c; \
-     gcc -Wall /sbin/su-exec.c -o/sbin/su-exec; \
-     chown root:root /sbin/su-exec; \
-     chmod 0755 /sbin/su-exec; \
-     rm /sbin/su-exec.c
+    curl -o /sbin/su-exec.c https://raw.githubusercontent.com/ncopa/su-exec/${SU_EXEC_VER}/su-exec.c; \
```
hjiawei, line 54: We can probably use `docker run --user UID:GID` to replace `su-exec` but that requires Calico Makefile changes. I would like to leave this for future improvements.

Dockerfile.amd64
```diff
@@ -34,125 +34,91 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
...
+    gcc -Wall /sbin/su-exec.c -o/sbin/su-exec; \
+    chown root:root /sbin/su-exec; \
+    chmod 0755 /sbin/su-exec; \
+    rm /sbin/su-exec.c
 
-# Install fossa for foss license checks
```
hjiawei, line 62: Doesn't seem to be used any more so removed.

Dockerfile.amd64
```diff
@@ -34,125 +34,91 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
...
-  go install gotest.tools/gotestsum@latest && \
-  rm -rf /go/src/* /root/.cache
-
 # Install ginkgo v2 as ginkgo2 and keep ginkgo v1 as ginkgo
-RUN mv /go/bin/ginkgo /go/bin/ginkgo1 && \
-  go install github.com/onsi/ginkgo/v2/ginkgo@v2.3.0 && \
-  mv /go/bin/ginkgo /go/bin/ginkgo2 && \
-  mv /go/bin/ginkgo1 /go/bin/ginkgo
+RUN \
+    go install github.com/onsi/ginkgo/v2/ginkgo@v2.9.2 && \
```
hjiawei, line 70: ginkgo v2 is upgraded to v2.9.2 and k8s.io libs to v0.25.8. Others are not changed (whitespace changes).

Dockerfile.amd64
```diff
@@ -34,125 +34,91 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
...
 # We can remove this once we update the Calico v3 APIs to use only types which are supported by the upstream controller-gen
 # tooling. Example: float, all the types in the numorstring package, etc.
 RUN wget -O ${GOPATH}/bin/controller-gen https://github.com/projectcalico/controller-tools/releases/download/calico-0.1/controller-gen && chmod +x ${GOPATH}/bin/controller-gen
 
 # Enable non-native runs on amd64 architecture hosts
-RUN for i in ${QEMU_ARCHS}; do curl -L https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar zxvf - -C /usr/bin; done
+RUN for i in ${QEMU_ARCHS}; do curl -sfL https://github.com/multiarch/qemu-user-static/releases/download/v${QEMU_VERSION}/qemu-${i}-static.tar.gz | tar xz -C /usr/bin; done
 RUN chmod +x /usr/bin/qemu-*
 
-# When running cross built binaries run-times will be auto-installed,
```
hjiawei, line 133: We have separate Dockerfiles for cross builds so we may not need multiarch std libs in amd64 image?

hjiawei:
Here is the Calico OSS main branch [build result](https://tigera.semaphoreci.com/workflows/d06298ea-c9d5-49df-a1de-6c447174d013?pipeline_id=ae82fe84-8de3-47dd-8fa3-63856eb46895). "Containerd - Windows FV" failure feels like a flake. Other results are similar to current go-build results.

hjiawei:
The compressed image size is down to 1.09 GB (from v1.77 GB in v0.83).

Dockerfile.amd64
```diff
@@ -34,125 +34,91 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
     apt-get install --no-install-recommends -y -t buster-backports \
         libbpf-dev linux-headers-amd64  && \
     apt-get install --no-install-recommends -y \
-        curl bash git openssh-client mercurial make wget util-linux file grep sed jq zip \
-        lsb-release software-properties-common binutils file iproute2 \
+        curl git openssh-client make wget util-linux file grep sed jq zip \
+        lsb-release software-properties-common binutils iproute2 \
         ca-certificates gcc mingw-w64 libc-dev bsdmainutils strace libpcap-dev \
```
hjiawei, line 39: Removing `mingw-w64` can save us another ~1.5GB uncompressed. Maybe this is another place for us to improve to split into a windows only build image.

Dockerfile.amd64
```diff
@@ -34,125 +34,91 @@ RUN echo 'APT::Default-Release "buster";' > /etc/apt/apt.conf.d/99defaultrelease
...
-    mv crane /usr/bin
+RUN curl -sfL https://github.com/google/go-containerregistry/releases/download/v0.4.1/go-containerregistry_Linux_x86_64.tar.gz | tar xz -C /usr/bin crane
 
 # Add bpftool for Felix UT/FV.
 COPY --from=bpftool /bpftool /usr/bin
 
 # Build ModSecurity for Dikastes.
-COPY scripts/modsec.sh /usr/local/bin/scripts/modsec.sh
-RUN /usr/local/bin/scripts/modsec.sh
+RUN git clone -b ${MODSEC_VERSION} --depth 1 --recurse-submodules --shallow-submodules https://github.com/SpiderLabs/ModSecurity.git /build && \
```
fasaxc, line 118: Did you try an enterprise build with this; modsec is only used in enterprise

hjiawei: Yes. The app-policy builds and UTs are green in enterprise.

rene-dekker (approved):
lgtm

PR #436, opened 2023-04-13
Make Tests stage depend on Push manifest stage

hjiawei:
This changeset makes "Tests" task depends on the "Push manifest" task. Calico uses multi-arch manifests in builds [1] but go-build doesn't push multi-arch manifests consistently.

When we build new changes in the main branch or a PR, the "Push manifest" task is running in parallel with "Tests" task. The "Tests" task will likely to fail because dockerhub doesn't have the manifests yet.

For an existing branch, like release branches v0.xy, new changes in branch won't recreate the manifests. This results different hashes for `calico/go-build:v0.xy` (older image) and `calico/go-build:v0.xy-arch` (newer image).

[1] https://github.com/projectcalico/calico/blob/master/lib.Makefile#L150

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -3,7 +3,7 @@ name: go-build
 agent:
   machine:
     type: e1-standard-2
-    os_image: ubuntu1804
+    os_image: ubuntu2004
 
 execution_time_limit:
   minutes: 60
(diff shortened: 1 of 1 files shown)


.semaphore/semaphore.yml
```diff
@@ -17,68 +17,64 @@ global_job_config:
...
+        - name: quay-robot-calico-and-semaphoreci
+        - name: docker
+      jobs:
+        - name: Push multi-arch manifest
+          commands:
+            - echo $DOCKER_TOKEN | docker login --username "$DOCKER_USER" --password-stdin
+            - echo $QUAY_TOKEN | docker login --username "$QUAY_USER" --password-stdin quay.io
+            - export BRANCH_NAME=$SEMAPHORE_GIT_BRANCH
+            - export VERSION=$BRANCH_NAME
+            - make push-manifest CONFIRM=true
```
fasaxc, line 55: Did you mean to remove the condition; we don't normally push images for PRs?

hjiawei: We probably do want to push the multi-arch manifest as well for PRs. Currently, we push images for PRs in the "Build image" stage (`calico/go-build:$branch-$arch`) but not updating manifests (`calico/go-build:$branch`). When running "Tests" stage, they are actually testing with an old hash because `calico/go-build:$branch` images are pulled.

Brian-McM: But this is overriding it for the branch with unmerged code, which I don't think is something we want to do. Two PRs running at the same time have a race condition in this case as well. 

I'd argue that if the PR is pushing images with the branch name then that's likely not right either. 

I'd say either don't push the images and the manifests and test them without having to upload them or upload them with the PR name. Former would preferred if possible.

hjiawei: Alright. My bad. the `SEMAPHORE_GIT_PR_NUMBER` condition check is back.

PR #442, opened 2023-05-17
calico/go-build s390x image on Dockerhub seems to be missing tools for testing

rposts:
While running Calico testcases on `s390x` arch it is observed that `calico/go-build:v0.85-s390x` does not contain few components when compared to `amd64`. This is also evident in the image files sizes for `s390x` (498.06MB) and `amd64` (1.08 GB).

This PR attempts to bring `s390x` image closer to `amd64`.

Since `Dockerfile.s390x` changes implemented by this PR require correct `calico/bpftool:v5.3-s390x` image, this issue must be also be addressed first: https://github.com/projectcalico/calico/issues/7654

Thanks.

### DIFF
--- a/Dockerfile.s390x
+++ b/Dockerfile.s390x
@@ -1,25 +1,28 @@
-FROM alpine:3.17 as qemu
+FROM calico/bpftool:v5.3-s390x as bpftool
+
+FROM debian:buster as qemu
+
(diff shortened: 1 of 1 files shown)


rposts:
Please let me know how we can proceed with this PR - Thanks!

hjiawei:
This PR should be addressed in https://github.com/projectcalico/go-build/pull/490. Let me know if anything is missing from current master.

hjiawei:
Close as completed in https://github.com/projectcalico/go-build/pull/490.

PR #459, opened 2023-07-13
Fix the ut's that don't work before an image is pushed.

rene-dekker:
Existing situation:
- When someone pushes a branch directly (or when a PR is merged) to https://github.com/projectcalico/go-build, CI/CD will run that pushes images.
- When someone forks this repo and creates a PR, the CI that runs will not push images.

The above is good and all, but it becomes problematic when the UT suite is dependent on images being pushed. Therefore the UTs are completely broken for any outside contributor who does not have the right privileges or connections to get images pushed.

This PR runs the UT suite directly after the amd64 image is built and removes the UT block. Other options explored:
- Building images again inside the UT. I found this a bit wasteful
- Caching images in sem is not an option, since the images are so large.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -20,37 +20,17 @@ global_job_config:
       - git fetch --unshallow
 
 blocks:
-# Run the full set of tests. Each job can potentially be run in parallel,
-# provided Semaphore has enough boxes available.
-- name: "Tests"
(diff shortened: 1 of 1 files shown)


fasaxc (approved):
Looks like a pragmatic change to me. Only thing we lose is the separate pass/fail for the UT job

PR #460, opened 2023-07-16
Install ping utility for bullseye

hjiawei:
The ping utility doesn't seem to be installed by default in debian bullseye. This fails egress-gateway UTs which depend on ping to exist in the system search path.


Behnam-Shobiri (approved):
lgtm

Behnam-Shobiri:
@hjiawei I think we need this in 0.87 branch as well

hjiawei:
> @hjiawei I think we need this in 0.87 branch as well

This is picked in https://github.com/projectcalico/go-build/pull/461.

PR #461, opened 2023-07-16
[v0.87] Install ping utility for bullseye

hjiawei:
The ping utility doesn't seem to be installed by default in debian bullseye. This fails egress-gateway UTs which depend on ping to exist in the system search path.

Pick https://github.com/projectcalico/go-build/pull/460 into v0.87 branch


Behnam-Shobiri (approved):
lgtm

PR #465, opened 2023-08-17
Update golang to 1.19.12

mgleung:
Update golang to 1.19.12


frozenprocess (approved):

frozenprocess (changes requested):
If we are going with bookworm then llvm should change to 15
https://github.com/projectcalico/go-build/blob/0f2fbce9c1a2bdfeeae3e59fcaabb542849385d9/Dockerfile.arm64#L48-L49
https://apt.llvm.org/bookworm/dists <- min is 15

frozenprocess (approved):

PR #469, opened 2023-09-07
[0.87] Update golang to 1.20.8

Behnam-Shobiri:
Update golang to 1.20.8
https://github.com/projectcalico/go-build/pull/468


rene-dekker (approved):
lgtm

PR #475, opened 2023-10-04
Update golangci-lint to v1.54.2 for compatibility with golang 1.21

Brian-McM:


hjiawei (approved):
Please pick to `v0.88` and `v0.89`.

PR #491, opened 2023-10-17
Update release tag calculation

danudey:
A simple update to release tag calculation.

The old regex would search for specifically `v\d*.\d*.\d*`, which would correctly extract a standard `vX.Y.Z` version. With our new `vX.Y.Z-a.b` versioning, this would mangle the intended version name.

The new regex searches for `v\d*.\d*.\d*`, then an optional `-.*`, and then includes a lookahead for `-<dev tag suffix>`. This causes the optional `-.*` to end before the dev tag suffix, wherever that is in the `git describe`, and should theoretically support any similar semver-compatible tags we might use in future.

Examples (for `DEV_TAG_SUFFIX=devel`)

| | v3.18.0-devel | v3.18.0-1.1-devel |
|----|--------|--------|
| Old | v3.18.0 | v3.18.0 |
| New | v3.18.0 | v3.18.0-1.1 |

This shouldn't cause any unforseen side effects, unless
1. DEV_TAG_SUFFIX is set incorrectly somewhere; or
2. DEV_TAG_SUFFIX is used incorrectly somewhere

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -889,7 +889,7 @@ fetch-all:
 # git-dev-tag retrieves the dev tag for the current commit (the one are dev images are tagged with).
 git-dev-tag = $(shell git describe --tags --long --always --abbrev=12 --match "*dev*")
 # git-release-tag-from-dev-tag get's the release version from the current commits dev tag.
-git-release-tag-from-dev-tag = $(shell echo $(call git-dev-tag) | grep -P -o "^v\d*.\d*.\d*")
+git-release-tag-from-dev-tag = $(shell echo $(call git-dev-tag) | grep -P -o "^v\d*.\d*.\d*(-.*)?(?=-$(DEV_TAG_SUFFIX))")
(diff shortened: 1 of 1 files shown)


hjiawei (approved):
Good for now but it might be better to change the middle part `.*` to be more specific and we probably don't need to use lookahead.

hjiawei:
We also have similar functions in Calico repo as well, for example, this [line](https://github.com/projectcalico/calico/blob/39cc89e0f520c39b2bc9b6e50f44df403c0b4e51/lib.Makefile#L1000). Are you going to change all of them?

danudey:
> We also have similar functions in Calico repo as well, for example, this [line](https://github.com/projectcalico/calico/blob/39cc89e0f520c39b2bc9b6e50f44df403c0b4e51/lib.Makefile#L1000). Are you going to change all of them?

Yes, definitely.

PR #495, opened 2023-11-08
[v0.88] update golang to 1.21.4

Behnam-Shobiri:
update golang to 1.21.4
https://github.com/projectcalico/go-build/pull/493


rene-dekker (approved):
lgtm

PR #501, opened 2023-11-30
[v0.90] Add mingw64-gcc package for Windows CGO builds

hjiawei:
Pick https://github.com/projectcalico/go-build/pull/500 into release v0.90 branch.


rene-dekker (approved):
lgtm

PR #502, opened 2023-11-30
Fix version number generation for v3.18.0-2.0

danudey:
Make version parsing more flexible to handle the new EP version scheme


rene-dekker (approved):
lgtm

PR #504, opened 2023-11-30
Fix version number generation for v3.18.0-2.0

danudey:
Make version parsing more flexible to handle the new EP version scheme


rene-dekker (approved):
lgtm

hjiawei:
Maybe not needed once we bump go-build version to v0.89 for all release repositories. This go-build version is too old and outdated dependencies may have unexpected behaviors.

PR #513, opened 2023-12-06
Create calico/base and calico/qemu-user-static common images

hjiawei:
This changeset creates `calico/base` and `calico/qemu-user-static` common base image for all supported architectures.


.semaphore/semaphore.yml
hjiawei, line 1: Mostly yaml reformat with minor name/description changes.

.semaphore/semaphore.yml
```diff
@@ -20,49 +20,76 @@ global_job_config:
...
+    dependencies: []
+    task:
+      secrets:
+        - name: docker
+      jobs:
+        - name: Build and push qemu image
+          commands:
+            - echo $DOCKER_TOKEN | docker login --username "$DOCKER_USER" --password-stdin
+            - make image-qemu ARCH=$TARGET_ARCH
+            - if [ -z "${SEMAPHORE_GIT_PR_NUMBER}" ]; then make push-qemu ARCH=$TARGET_ARCH CONFIRM=true; fi
```
coutinhop, line 75: should this be split into separate 'build' and 'push' (with push only running for branches, like the 'Push multi-arch manifests' job)?

Makefile.common
```diff
@@ -846,10 +846,6 @@ sub-manifest-%:
 	$(DOCKER) manifest create $(call unescapefs,$*):$(IMAGETAG) $(addprefix --amend ,$(addprefix $(call unescapefs,$*):$(IMAGETAG)-,$(VALIDARCHES)))
 	$(DOCKER) manifest push --purge $(call unescapefs,$*):$(IMAGETAG)
 
-push-manifests-with-tag: var-require-one-of-CONFIRM-DRYRUN var-require-all-BRANCH_NAME
```
coutinhop, line 849: Do we use this 'Makefile.common' elsewhere? (I mean, do we curl/wget it from this repo? I forget which repo we do that from, but it could very well be this one...) If we don't, it doesn't seem to be used anywhere in this repo and could be removed... Otherwise, can you clarify why we're removing these lines?

hjiawei: This is added in https://github.com/projectcalico/go-build/pull/516 which picks a change from private. However, the `cd-common` here is a bit different which includes push-manifests. It is not used by any other repositories at the moment.

.semaphore/semaphore.yml
hjiawei: The build is done in `make image-qemu` (one line above). It runs in every PR. The push will only run when SEMAPHORE_GIT_PR_NUMBER is empty (after merging to master).

coutinhop (approved):

PR #515, opened 2023-12-11
Remove glibc-static dependency

hjiawei:
We can't statically link glibc so remove it as a dependency. This change also cleans up Makefile.


rene-dekker (approved):
lgtm

PR #517, opened 2023-12-12
[v0.90] Add push-manifests-with-tag target in common Makefile

hjiawei:
This changset adds `push-manifests-with-tag` target in the common Makefile to push multi-arch manifests. It creates multi-arch manifests with branch name and git hash.


rene-dekker (approved):
lgtm

PR #519, opened 2024-01-03
Use host native calico/go-build

hjiawei:
This changeset enforces host native calico/go-build and removes outdated version definitions.


hjiawei:
Included in https://github.com/projectcalico/go-build/pull/520.

PR #520, opened 2024-01-03
Remove NONMANIFEST_REGISTRIES that causes empty MANIFEST_REGISTRIES

hjiawei:
This change removes NONMANIFEST_REGISTRIES variable defined in common Makefile that causes MANIFEST_REGISTRIES to be empty. This affects all repositories that include this common makefile. Calico components enabled multi-arch builds and all registries seem to support multi-arch manifests now. It also removes outdated version definitions.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -92,9 +92,7 @@ filter-registry ?= $(if $(filter-out $(1),$(DOCKERHUB_REGISTRY)),$(1)/)
 # Convenience function to get the first dev image repo in the list.
 DEV_REGISTRY ?= $(firstword $(DEV_REGISTRIES))
 
-# remove from the list to push to manifest any registries that do not support multi-arch
-NONMANIFEST_REGISTRIES      ?=
-MANIFEST_REGISTRIES         ?= $(DEV_REGISTRIES:$(NONMANIFEST_REGISTRIES)%=)
+MANIFEST_REGISTRIES ?= $(DEV_REGISTRIES)
 
 PUSH_MANIFEST_IMAGES := $(foreach registry,$(MANIFEST_REGISTRIES),$(foreach image,$(BUILD_IMAGES),$(call filter-registry,$(registry))$(image)))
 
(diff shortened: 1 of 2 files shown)


Makefile.common
```diff
@@ -132,14 +130,8 @@ endif
...
 GO_BUILD_IMAGE ?= calico/go-build
-CALICO_BUILD    = $(GO_BUILD_IMAGE):$(GO_BUILD_VER)
-
-BIRD_VERSION=v0.3.1
-COREDNS_VERSION=1.5.2
-ETCD_VERSION=v3.5.1
-K8S_VERSION=v1.23.2
-KUBECTL_VERSION=v1.23.2
-PROTOC_VER=v0.1
+CALICO_BUILD    = $(GO_BUILD_IMAGE):$(GO_BUILD_VER)-$(BUILDARCH)
```
hjiawei, line 133: This change is to match the comment above to "use the *host* architecture". When `--platform` is specified in docker cli, it may find the multi-arch image to run cross builds. Using `-$(BUILDARCH)` will force it to use the host native calico/go-build.

Makefile.common
```diff
@@ -92,9 +92,7 @@ filter-registry ?= $(if $(filter-out $(1),$(DOCKERHUB_REGISTRY)),$(1)/)
 # Convenience function to get the first dev image repo in the list.
 DEV_REGISTRY ?= $(firstword $(DEV_REGISTRIES))
 
-# remove from the list to push to manifest any registries that do not support multi-arch
-NONMANIFEST_REGISTRIES      ?=
-MANIFEST_REGISTRIES         ?= $(DEV_REGISTRIES:$(NONMANIFEST_REGISTRIES)%=)
```
hjiawei, line 97: After this substitution, `MANIFEST_REGISTRIES` becomes empty so multi-arch manifest won't be pushed. In Calico, we [override it](https://github.com/projectcalico/calico/blob/master/lib.Makefile#L104). As Calico Enterprise requires multi-arch manifest to work properly on ARM64 cluster, we probably will never push images to a registry that doesn't support multi-arch.

rene-dekker (approved):
lgtm

PR #521, opened 2024-01-04
[v0.90] Remove NONMANIFEST_REGISTRIES that causes empty MANIFEST_REGISTRIES

hjiawei:
This change removes NONMANIFEST_REGISTRIES variable defined in common Makefile that causes MANIFEST_REGISTRIES to be empty. This affects all repositories that include this common makefile. Calico components enabled multi-arch builds and all registries seem to support multi-arch manifests now. It also removes outdated version definitions.

Pick https://github.com/projectcalico/go-build/pull/520 into v0.90 branch.


rene-dekker (approved):
lgtm

PR #526, opened 2024-01-09
[0.89] Update golang to 1.21.6 & update Alpine to 3.19

Behnam-Shobiri:
Update golang to 1.21.6
Update Alpine to 3.19
https://github.com/projectcalico/go-build/pull/525


rene-dekker (approved):
lgtm

tmjd:
We shouldn't be backporting any changes for go-build because this effectively changes the tagged image we use to build with, resulting in un-reproducable older builds.

PR #528, opened 2024-01-09
[v0.87] Update golang to 1.20.13  & update Alpine to 3.19

Behnam-Shobiri:
Update golang to 1.20.13
Update Alpine to 3.19
https://github.com/projectcalico/go-build/pull/525


rene-dekker (approved):
lgtm

PR #529, opened 2024-01-09
Copy proper dynamic loader symbolic links for all supported arches

hjiawei:
This change copies dynamic loader symbolic links into calico/base image. As they varies among different arches, they are summarized as follows:

| arch    | loader location   | symblink in /lib64    | symblink in /lib      |
| ------- | ----------------- | --------------------- | --------------------- |
| amd64   | /lib64/ld-2.28.so | ld-linux-x86-64.so.2  | N/A                   |
| arm64   | /lib64/ld-2.28.so | ld-linux-aarch64.so.1 | ld-linux-aarch64.so.1 |
| ppc64le | /lib64/ld-2.28.so | ld64.so.2             | N/A                   |
| s390x   | /lib64/ld-2.28.so | ld64.so.1             | ld64.so.1             |

Note that the loader name and symbolic links will change in UBI9 so we need to revisit this when we upgrade.

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -2,30 +2,45 @@ FROM --platform=linux/amd64 calico/qemu-user-static:latest as qemu
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest as ubi
 
+ARG LDSONAME
+
 COPY --from=qemu /usr/bin/qemu-*-static /usr/bin/
 
 RUN microdnf upgrade -y
 
-FROM scratch as source
+# Prepare a rootfs for necessary files from UBI.
+# Symbolic links are preserved.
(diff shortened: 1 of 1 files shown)


base/Dockerfile
```diff
@@ -2,30 +2,45 @@ FROM --platform=linux/amd64 calico/qemu-user-static:latest as qemu
...
 
 RUN microdnf upgrade -y
 
-FROM scratch as source
+# Prepare a rootfs for necessary files from UBI.
+# Symbolic links are preserved.
+RUN mkdir -p /rootfs/lib64 /rootfs/etc
 
-ARG LDSONAME
+# Copy dyanmic loader and symbolic links.
```
rene-dekker, line 15: typo: dynamic.

rene-dekker (approved):
lgtm

PR #530, opened 2024-01-25
Fix entrypoint extra group add

hjiawei:


entrypoint.sh
```diff
@@ -9,26 +9,25 @@ if [ "${RUN_AS_ROOT}" = "true" ]; then
...
 
 export HOME=/home/user
 
 if [ -n "$EXTRA_GROUP_ID" ]; then
-  echo "Adding user to additional GID : $EXTRA_GROUP_ID" 1>&2
+  echo "Adding user to additional GID: $EXTRA_GROUP_ID" 1>&2
   # Adding the group can fail if it already exists.
-  if addgroup --gid "$EXTRA_GROUP_ID" group; then
-    adduser user group
+  if groupadd --gid "$EXTRA_GROUP_ID" group; then
```
hjiawei, line 26: `addgroup` wrapper doesn't exist in RHEL/UBI so change to more primitive `groupadd` calls.

entrypoint.sh
```diff
@@ -9,26 +9,25 @@ if [ "${RUN_AS_ROOT}" = "true" ]; then
...
+    usermod -a -G group user
```
hjiawei, line 27: At this point, `user` should exist already so use `usermod` to append `user` to new groups. However, this `EXTRA_GROUP_ID` doesn't seem to be used or used correctly in Calico code base so we will review and may remove it soon.

rene-dekker (approved):
lgtm

PR #539, opened 2024-01-30
Remove $(double_quote) from crane cmds [master]

danudey:


Makefile.common
```diff
@@ -388,7 +388,7 @@ git-commit:
 # different implementation.
 ###############################################################################
 
-CRANE_CMD         = docker run -v $(DOCKER_CONFIG):/home/user/.docker/config.json $(CALICO_BUILD) crane
+CRANE_CMD         = docker run -e LOCAL_USER_ID=$(id -u) -v $(DOCKER_CONFIG):/home/user/.docker/config.json $(CALICO_BUILD) crane
```
hjiawei, line 391: Use `$(shell id- u)`.

danudey: Switched to existing LOCAL_USER_ID variable instead

hjiawei (approved):

PR #544, opened 2024-02-15
[v0.90] Add xz system utility

hjiawei:
Pick https://github.com/projectcalico/go-build/pull/543 into v0.90 branch.


rene-dekker (approved):
lgtm

PR #546, opened 2024-03-05
Remove push-manifests from cd-common target

hjiawei:
This change removes push-manifests from cd-comm into a separate push-manifest-with-tag target. When building native on different platforms, we don't want cd-common to publish multi-arch manifests because not all images are pushed.


hjiawei:
We have migrated most of our microrepos into monorepo so this change isn't needed anymore.

PR #547, opened 2024-03-06
Update golang to the latest patch version

rene-dekker:


hjiawei (approved):

hjiawei:
Have to pin `openapi-gen` and `deepcopy-gen` to a release version for armv7, ppc64le, and s390x.

PR #551, opened 2024-04-03
Build and add `lipo` from Apple's cctools for making universal binaries

danudey:
We currently build amd64 and arm64 binaries for macOS but release them separately. Using Apple's `lipo` tool, we can build a universal binary which will run natively on both arm64 and amd64 macs.

This would allow us to publish just one binary, and allow users to download just one binary, without having to worry about which is which (and it's easy to mistake 'amd64' and 'arm64' and download the wrong one by accident).

Currently this only builds the `lipo` binary for inclusion into the amd64 docker image. It does build successfully on the arm64 docker image as well, but I'm not sure if it's important to build it on both.

On my system it adds about 20 seconds to the image build time, which is about half of what it takes to download `kube-apiserver`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -130,6 +130,21 @@ RUN set -eux; \
     gcc -Wall -O2 /tmp/su-exec.c -o /usr/bin/su-exec; \
     rm -f /tmp/su-exec.c
 
+# Build lipo (for creating macOS universal binaries)
+
+RUN --mount=type=tmpfs,target=/tmp/cctools-port \
+    set -eux; \
+    if [ "${TARGETARCH}" = "amd64" ]; then \
+    dnf install -y libdispatch && \
+    git clone https://github.com/tpoechtrager/cctools-port /tmp/cctools-port && \
+        cd /tmp/cctools-port/cctools && \
+        ./configure && \
+        make -j12 -C libmacho && \
+        make -j12 -C libstuff && \
+        make -j12 -C misc lipo && \
+        cp misc/lipo /usr/local/bin; \
+    fi
+
 # Install Go utilities
 
 # controller-gen is used for generating CRD files.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -130,6 +130,21 @@ RUN set -eux; \
...
+# Build lipo (for creating macOS universal binaries)
+
+RUN --mount=type=tmpfs,target=/tmp/cctools-port \
+    set -eux; \
+    if [ "${TARGETARCH}" = "amd64" ]; then \
+    dnf install -y libdispatch && \
+    git clone https://github.com/tpoechtrager/cctools-port /tmp/cctools-port && \
+        cd /tmp/cctools-port/cctools && \
+        ./configure && \
+        make -j12 -C libmacho && \
```
hjiawei, line 142: Suggested -j4. We build on a 4-core semaphore VM. -j12 will over schedule make progress and may cause extra scheduling/synchronization overheads.

Dockerfile
```diff
@@ -130,6 +130,21 @@ RUN set -eux; \
...
+        make -j12 -C libstuff && \
+        make -j12 -C misc lipo && \
+        cp misc/lipo /usr/local/bin; \
+    fi
```
hjiawei, line 146: Cleanup add intermediate build objects. It is better to build in a temporary folder and delete it after binaries are copied.

hjiawei (changes requested):
Why do we build it from source? Is it not in any UBI or Rocky repositories? Also see comments online.

Dockerfile
danudey: Will fix this, thanks for the catch!

Dockerfile
danudey: On line 135 we start this RUN with `RUN --mount=type=tmpfs,target=/tmp/cctools-port`, which creates a tmpfs directory at the target location. We build there, and the tmpfs is cleaned up automatically when this build stage is over so we don't have to worry about manually cleaning up.

danudey:
> Why do we build it from source? Is it not in any UBI or Rocky repositories? Also see comments online.

I have not been able to find a build of these tools in any binary repositories (neither for UBI/Rocky nor Ubuntu), nor a pre-built Docker image for the project. Given that and given how quickly we can build it, I opted to compile a version for our use.

danudey:
Nevermind me, apparently LLVM has included `lipo` as a utility for... a long time.

PR #553, opened 2024-04-11
Fix generation of image tags based on dev tags

pasanw:
Currently, the generation of image tags yields the incorrect tag when the current commit has the dev tag on it.

This is because https://github.com/projectcalico/go-build/pull/516 removed the usage of the `--long` flag when generating the image tag from `git describe`.

This PR uses `git-dev-tag` instead of `GIT_VERSION` to restore the long flag.


rene-dekker (approved):
lgtm

PR #561, opened 2024-05-07
[v0.90] update golang 1.21.10

Behnam-Shobiri:
update golang 1.21.10 
https://github.com/projectcalico/go-build/pull/559


rene-dekker:
@Behnam-Shobiri  Calico can no longer be built with 1.21, so the automation fails. I added a commit so it runs against calico 3.27

PR #562, opened 2024-05-22
Rev k8s to 1.29.4

aaaaaaaalex:


Makefile
```diff
@@ -105,6 +105,7 @@ push: image
 	docker push $(GOBUILD_ARCH_IMAGE)
 	# to handle default case, because quay.io does not support manifest yet
 ifeq ($(ARCH),amd64)
+	docker tag $(GOBUILD_ARCH_IMAGE) $(GOBUILD_IMAGE)
```
hjiawei, line 108: Why do you need this change?

aaaaaaaalex: This may not be the ideal fix for the failure, so Im open to suggestions, but:
[A CI run is failing](https://tigera.semaphoreci.com/jobs/559a9212-fcbb-47cd-a047-d4a0c723498b) with error:
```
docker: Error response from daemon: manifest for calico/go-build:rev-k8s-to-1-29-4 not found: manifest unknown: manifest unknown.
```

So, this was a short attempt to fix that, though the change did not appear to get used in the CI following this commit :thinking:

hjiawei: It needs to be a [multi-arch manifest](https://github.com/projectcalico/go-build/blob/master/.semaphore/semaphore.yml#L95); not necessarily a retagged amd46 image. For your work in https://github.com/projectcalico/calico/pull/8805, you probably can get away with updating go-build. The current version (v0.91) should be good to use. I believe the k8s libs/binaries in the build container is mostly for UT/FVs.

aaaaaaaalex: That's fair enough, but what do you suggest is the fix then, for the missing image in the CI? Or maybe a better question is why is it failing? :smile:

aaaaaaaalex: > I believe the k8s libs/binaries in the build container is mostly for UT/FVs.

That sounds plausible, but Im seeing quite a few failures in the tests as a result of revving the libs. Should this be cause for concern? Im currently in the process of deploying a real test cluster with Calico pinned to k8s1.29 to see if the issues arise outside of just the tests

hjiawei: > for the missing image in the CI

You mean the projectcalico/calico CI failure or this go-build CI failure? For calico, I used to push images+multi-arch manifest to my own dockerhub account and use that. For the go-build CI, I am unable to fetch logs for this build.

aaaaaaaalex: The latest CI run appears to be bugged - will push another commit and hopefully thats visible

PR #563, opened 2024-05-28
update next version to account for EP releases

radTuti:
This updates the next version tag to account for EP releases.


danudey (approved):
lgtm

caseydavenport (approved):

PR #564, opened 2024-06-01
Install necessary toolchain for ppc64le and s390x

hjiawei:
This change makes ppc64le and s390x toolchain parity with amd64 and arm64. RockyLinux 8 doesn't support ppc64le nor s390x so the additional toolchain repository is switched to AlmaLinux.


hjiawei:
This CI failure is due to clang v16 being updated to v17 from RHEL 8.10. We will address this separately.

Behnam-Shobiri (approved):

PR #573, opened 2024-06-14
Test local tag and ci build

hjiawei:


hjiawei:
Tests are green at https://tigera.semaphoreci.com/workflows/4aaf612e-d4e5-4b8b-ac7f-4246fe732e50?pipeline_id=d316e608-1f13-4615-b9d4-e71aa433024b so the retag works as expected.

PR #587, opened 2024-09-15
Bump golang to v1.23.1 and update utilities

hjiawei:
This change bumps golang to v1.23.1 and updates some utilities to the latest release.


Behnam-Shobiri (approved):
lgtm

PR #596, opened 2024-10-16
[semvalidator] update skip dirs usage and add skipping files

radTuti:


radTuti:
closing as this is stale

PR #601, opened 2024-11-13
Use calico branch v3.27 for go-build v0.92

coutinhop:


rene-dekker (approved):
From what I understand, versions newer than 0.27 use a version of clang that this branch wouldn't be able to build. Therefore, lgtm

PR #609, opened 2024-11-22
[v0.89] Upgrade golang to v1.22.9 and k8s to v1.27.16

coutinhop:
Also, stringer to v0.27.0 and bpftool to v7.4.0

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -1,11 +1,11 @@
-FROM calico/bpftool:v5.3-amd64 as bpftool
+FROM calico/bpftool:v7.4.0 AS bpftool
 
-FROM golang:1.21.13-bullseye
+FROM golang:1.22.9-bullseye
 
 LABEL maintainer="Shaun Crampton <shaun@projectcalico.org>"
 
 ARG GO_LINT_VERSION=v1.54.2
-ARG K8S_VERSION=v1.26.14
+ARG K8S_VERSION=v1.27.16
 ARG LLVM_VERSION=15
 ARG MANIFEST_TOOL_VERSION=v1.0.2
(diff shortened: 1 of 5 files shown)


Dockerfile.amd64
```diff
@@ -40,6 +40,8 @@ RUN apt-get -y update && apt-get -y upgrade && \
 
 RUN curl -sfL https://apt.llvm.org/llvm.sh | bash -s -- ${LLVM_VERSION} && \
     apt-get install clang-${LLVM_VERSION}
+RUN ln -s /usr/bin/clang-${LLVM_VERSION} /usr/bin/clang
+RUN ln -s /usr/bin/llc-${LLVM_VERSION} /usr/bin/llc
```
hjiawei, line 44: We probably don't need these two links if we can revert the changes in https://github.com/projectcalico/calico/pull/9515/files#diff-d835d785042c5c950fc9cb0c37964a276b4edc85f2f896461db9853eb325ae88R51-R52.

Dockerfile.arm64
```diff
@@ -44,6 +44,8 @@ RUN apt-get update && apt-get -y upgrade && \
 
 RUN curl -sfL https://apt.llvm.org/llvm.sh | bash -s -- ${LLVM_VERSION} && \
     apt-get install clang-${LLVM_VERSION}
+RUN ln -s /usr/bin/clang-${LLVM_VERSION} /usr/bin/clang
+RUN ln -s /usr/bin/llc-${LLVM_VERSION} /usr/bin/llc
```
hjiawei, line 48: Same comments in the amd64 Dockerfile.

hjiawei (approved):

PR #611, opened 2024-12-03
Tag calico/go-build with compiler versions

hjiawei:
This change tags calico/go-build container image with more meaningful compiler versions. The new tag includes golang, clang, and kubernetes versions. If any of the three versions are updated, the new github action will create a branch with the same tag name. The new branch will then trigger semaphore builds and push images to dockerhub.


generate-version-tag.sh
```diff
@@ -0,0 +1,18 @@
...
+
+golang_ver=$(yq -r .golang.version "$ver_file")
+k8s_ver=$(yq -r .kubernetes.version "$ver_file")
+llvm_ver=$(yq -r .llvm.version "$ver_file")
+
+if [[ -z $golang_ver ]] || [[ -z $k8s_ver ]] || [[ -z $llvm_ver ]]; then
+    exit 1
+fi
+
+echo "${golang_ver}-llvm${llvm_ver}-k8s${k8s_ver}"
```
hjiawei, line 18: Returns, for example, `1.23.3-llvm18.1.8-k8s1.30.5`.

README.md
```diff
@@ -65,28 +62,6 @@ make register
 
 After the above registration, your system can handle other-architecture binaries. The above registration uses the first method, since _all_ kernels that support `binfmt` support this method, while only kernels from version 4.8+ support the latter. While docker-for-mac and docker-for-windows both use supporting kernels, almost every CI-as-a-service does not.
 
-## Using binfmt in other Calico projects
```
hjiawei, line 68: This section seems to be irrelevant to `calico/go-build`. These instructions are already implemented in other Calico components.

Makefile
```diff
@@ -1,173 +1,20 @@
...
-	$(MAKE) push-qemu
-
-.PHONY: push-manifest
-push-manifest:
-	# Docker login to hub.docker.com required before running this target as we are using $(HOME)/.docker/config.json holds the docker login credentials
-	docker manifest create $(GOBUILD_IMAGE) $(addprefix --amend ,$(addprefix $(GOBUILD_IMAGE)-,$(ARCHES)))
-	docker manifest push --purge $(GOBUILD_IMAGE)
-	docker manifest create $(BASE_IMAGE) $(addprefix --amend ,$(addprefix $(BASE_IMAGE)-,$(ARCHES)))
-	docker manifest push --purge $(BASE_IMAGE)
+image:
```
danudey, line 8: Should this be PHONY as well?

Makefile.common
```diff
@@ -684,7 +685,7 @@ semaphore-run-workflow:
 
 # This is a helpful wrapper of the semaphore-run-workflow target to run the update_pins workflow file for a project.
 semaphore-run-auto-pin-update-workflow:
-	SEMAPHORE_WORKFLOW_FILE=update_pins.yml $(MAKE) semaphore-run-workflow
+	$(MAKE) semaphore-run-workflow
```
danudey, line 688: Is there a reason this needs to be a separate execution of make rather than just a dependent rule?

hack/generate-go-branch-name.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
```
danudey, line 3: I always prefer to `set -eu` so that we can get an error on undefined variables. Saves me a lot of bugs. Not necessary but highly recommended.

hack/generate-go-branch-name.sh
```diff
@@ -0,0 +1,31 @@
...
+        exit 1
+        ;;
+    *)
+        echo "invalid argument -$OPTARG" >&2
+        exit 1
+        ;;
+    esac
+done
+
+golang_ver=$(yq -r .golang.version "$ver_file")
```
danudey, line 23: Not sure if I missed this, but do we already guarantee somewhere else that we have `yq` in our path? I've made this assumption before and broken things.

hack/generate-version-tag-name.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -e
```
danudey, line 3: As above, `set -eu` recommended.

danudey (approved):

Makefile
hjiawei: It is a PHONY target (up at line 7).

hack/generate-go-branch-name.sh
hjiawei: Thanks. The set flag is updated to `-eu`.

hack/generate-version-tag-name.sh
hjiawei: Thanks. The set flag is updated to `-eu`.

hack/generate-go-branch-name.sh
hjiawei: I think semaphore ubuntu2204 build machines do install yq v4 by default. I don't think we need to download a different copy or use a containerized version.

PR #618, opened 2024-12-18
add CLEANUPLEVEL arg for image-base

aaaaaaaalex:
`make BASE_CLEANUPLEVEL=unstripped image-base` Generates an image which is scannable,

e.g. `trivy image calico/base:latest-amd64`.

`make BASE_CLEANUPLEVEL=stripped image-base` , or just `make image-base` generates the usual image with clean FS.

Make/Docker arg `UBIBASE` allows injection of a different image to build from.

Arg `(BASE_)PKGMAN` allows adjustment of pkgmanager binary name - useful if say, using DNF rather than MicroDNF. Necessary when using non-minimal UBI.

### DIFF
--- a/images/calico-base/Dockerfile
+++ b/images/calico-base/Dockerfile
@@ -4,7 +4,7 @@ ARG BASE
 
 FROM --platform=linux/amd64 calico/qemu-user-static:latest AS qemu
 
-FROM ${BASE} AS ubi
+FROM ${BASE} AS source-unstripped
 
 ARG LDSONAME
 
@@ -43,21 +43,15 @@ RUN cp /etc/nsswitch.conf /rootfs/etc/nsswitch.conf
 RUN cp /etc/os-release /rootfs/etc/os-release
 
 # Stripped image.
-FROM scratch AS intermediate-stripped
-
-COPY --from=ubi /rootfs /
-
+FROM scratch AS source-stripped
+COPY --from=source-unstripped /rootfs /
 # tmp.tar has a /tmp with the correct permissions 01777.
 ADD tmp.tar /
-
 COPY licenses /licenses/
 
-FROM scratch AS source-stripped
-COPY --from=intermediate-stripped / /
-
 # Unstripped image.
-FROM ${BASE} AS source-unstripped
-COPY --from=ubi / /
+FROM source-unstripped
+# Remove the prepped rootfs for the unstripped img.
 RUN rm -rf /rootfs
 
 # Choose which image is shipped.
--- a/images/calico-base/Dockerfile
+++ b/images/calico-base/Dockerfile
@@ -1,13 +1,17 @@
+# CLEANUPLEVEL defines the level of FS stripping we apply to the generated image.
+ARG CLEANUP_LEVEL
+ARG BASE
+
 FROM --platform=linux/amd64 calico/qemu-user-static:latest AS qemu
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS ubi
+FROM ${BASE} AS ubi
 
 ARG LDSONAME
 
 COPY --from=qemu /usr/bin/qemu-*-static /usr/bin/
 
-RUN microdnf upgrade -y
-
+ARG PKGMAN
+RUN ${PKGMAN} upgrade -y
 # Prepare a rootfs for necessary files from UBI.
 # Symbolic links are preserved.
 RUN mkdir -p /rootfs/lib64 /rootfs/etc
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -57,9 +57,17 @@ BASE ?= calico/base
 BASE_IMAGE ?= $(BASE):latest
 BASE_ARCH_IMAGE ?= $(BASE_IMAGE)-$(ARCH)
 
+
 QEMU ?= calico/qemu-user-static
 QEMU_IMAGE ?= $(QEMU):latest
 
+# Base-image we'll use to build calico/base.
+UBIBASE ?= registry.access.redhat.com/ubi8/ubi-minimal:latest
```
fasaxc, line 65:
```suggestion
BASE_BASE_IMAGE ?= registry.access.redhat.com/ubi8/ubi-minimal:latest
```
Bit of a stutter, but would align with the args below?

Makefile
```diff
@@ -57,9 +57,17 @@ BASE ?= calico/base
...
+# The level of cleanup we perform on the calico/base image. One-of: stripped, unstripped.
+BASE_CLEANUPLEVEL ?= stripped
```
fasaxc, line 67:
```suggestion
BASE_CLEANUP_LEVEL ?= stripped
```
I'd split those words

base/Dockerfile
```diff
@@ -38,15 +44,21 @@ RUN cp /etc/nsswitch.conf /rootfs/etc/nsswitch.conf
 # Copy base image release info.
 RUN cp /etc/os-release /rootfs/etc/os-release
 
-FROM scratch AS source
 
+# Stripped image.
+FROM scratch AS intermediate-stripped
```
fasaxc, line 49: Is this intermediate one needed?  Seems to duplicate source-stripped (or source-stripped could pull from ubi?

No licenses copied into source-unstripped?  Is that because they're already there?

Worth comments explaining these details so next maintainer doesn't need to ask.

aaaaaaaalex: Re: intermediate-stripped..
Prior to my changes, we build up the `scratch` image, and then the very bottom line says:

```
FROM scratch
COPY --from=source / /
```

I don't really understand this line, but it's why I've now got an `intermediate-stripped` and `source-stripped`. `intermediate-stripped` was called `source`, and `source-stripped` was the final step that I didn't fully see the point of, but tried to preserve.

I can ditch that last copy to a redundant scratch if you're confident it _is_ redundant!

hjiawei:
I once tried the dnf [`--installroot` flag](https://dnf.readthedocs.io/en/latest/command_ref.html) and installed glibc only. It gives an ~4MB image and also scannable. It ships way less dependencies then ubi-minimal (size similar to alpine).

images/calico-base/Dockerfile
```diff
@@ -47,6 +52,13 @@ ADD tmp.tar /
...
 COPY licenses /licenses/
 
-FROM scratch
+FROM scratch AS source-stripped
+COPY --from=intermediate-stripped / /
+
+# Unstripped image.
+FROM ${BASE} AS source-unstripped
+COPY --from=ubi / /
+RUN rm -rf /rootfs
```
hjiawei, line 61: This step results an extra layer to the source-unstripped image and wasted space if not squashed.

aaaaaaaalex: IMO I think that's fine - we dont intend to ship unstripped images, and anyone who want's to build an unstripped image themselves can opt to squash.

I have cleaned up these lines a bit anyway though, and removed what I believe are redundant layers. Does this help?

hjiawei:
Do we still need this change once https://github.com/projectcalico/calico/pull/9678 is merged?

aaaaaaaalex:
@hjiawei yep

hjiawei:
> @hjiawei yep

I am not following. With your change in https://github.com/projectcalico/calico/pull/9678, is it true that a user can run the following command to switch to a different base. Is this essentially what you are doing here?

```bash
CALICO_BASE=registry.access.redhat.com/ubi8/ubi-minimal:latest make image
```

images/calico-base/Dockerfile
fasaxc: I think we've ended up with qemu in the image too, which we don't want and maybe we now have literally nothing of value in the unstripped image (I thought we needed to add /tmp and the licenses, but if not, perhaps we should derive it from the BASE with no changes?

images/calico-base/Dockerfile
```diff
@@ -1,13 +1,17 @@
...
 
 COPY --from=qemu /usr/bin/qemu-*-static /usr/bin/
 
-RUN microdnf upgrade -y
-
+ARG PKGMAN
+RUN ${PKGMAN} upgrade -y
 # Prepare a rootfs for necessary files from UBI.
 # Symbolic links are preserved.
 RUN mkdir -p /rootfs/lib64 /rootfs/etc
```
fasaxc, line 17: I think all this rootfs building (and the qemu bit) should happen in an intermediate image that's _not_ a base for the unstripped one.

fasaxc (changes requested):

PR #636, opened 2025-01-17
Split calico go-build image publish into parallel pipelines

hjiawei:

PR #637, opened 2025-01-28
Fix calico/base multi-arch manifest push

hjiawei:
This change moves the slightly different `cd-common` implementations to each `calico-base-cd` and `calico-go-build-cd` target. `calico/base` multi-arch manifest isn't created properly after the parallel build changes in https://github.com/projectcalico/go-build/pull/636.


rene-dekker (approved):
lgtm

PR #639, opened 2025-02-02
[v0.95] Add protobuf-devel to calico/go-build and bump protoc-gen-go

hjiawei:
Install protobuf-devel package include protobuf well-known types like timestamp.proto. We need them to build felixbackend proto.


danudey (approved):
lgtm

PR #652, opened 2025-02-16
Build calico/base from UBI 8 and 9 and tag with timestamp

hjiawei:
This change builds `calico/base` from both UBI 8 and 9. They are also tagged with Unix timestamp instead of `latest`. This change also registers and installs qemu from `tonistiigi/binfmt` so that we don't need to maintain our own versions.


.semaphore/promotions/qemu-user-static.yml
hjiawei, line 1: Reference Calico PR https://github.com/projectcalico/calico/pull/9859. We don't need to maintain our own qemu static copies.

rene-dekker (approved):
lgtm

PR #654, opened 2025-02-24
[EV-5623] Upgrade go 1.23.6 and k8s 1.29.14

dimitri-nicolo:
* Upgrade `go` to `v1.23.6`
* Upgrade `k8s` to `v1.29.14`

Hash versions can be referenced:
https://go.dev/dl/


rene-dekker (approved):
lgtm

PR #670, opened 2025-04-07
Update calico/base image to fix issue with pod2demon-flexvol image on s390x

vivkong:
The `flexvol-driver` container on s390x crashed with the following error:

```
kubectl logs calico-node-n4htr -n calico-system -c flexvol-driver
exec /usr/local/bin/flexvol.sh: no such file or directory
```

The problem seems to be that `/lib/ld64.so.1` is missing from the s390x `calico/base` image which the `calico/pod2daemon-flexvol` image is based on.  Once `/lib/ld64.so.1` was added we were able to start Calico.

### DIFF
--- a/images/calico-base/Dockerfile.ubi8
+++ b/images/calico-base/Dockerfile.ubi8
@@ -13,7 +13,7 @@ RUN cp /lib64/ld-2.28.so /rootfs/lib64/ld-2.28.so
 RUN set -eux; \
     cp -a /lib64/${LDSONAME} /rootfs/lib64/${LDSONAME}; \
     if [ -f /lib/${LDSONAME} ]; then \
-        mkdir -p /rootfs/lib && cp -a /lib/${LDSONAME} /rootfs/lib/${LDSONAME}; \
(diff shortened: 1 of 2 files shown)


images/calico-base/Dockerfile.ubi8
```diff
@@ -13,7 +13,7 @@ RUN cp /lib64/ld-2.28.so /rootfs/lib64/ld-2.28.so
 RUN set -eux; \
     cp -a /lib64/${LDSONAME} /rootfs/lib64/${LDSONAME}; \
     if [ -f /lib/${LDSONAME} ]; then \
-        mkdir -p /rootfs/lib && cp -a /lib/${LDSONAME} /rootfs/lib/${LDSONAME}; \
+        mkdir -p /rootfs/lib && cp /lib/${LDSONAME} /rootfs/lib/${LDSONAME}; \
```
hjiawei, line 16: Thanks for identifying this issue. Updating the symlink at `/lib/ld64.so.1 → /usr/lib64/ld64.so.1` to point to `../lib64/ld64.so.1` may help resolve the s390x problem. We use `cp -a` to preserve symbolic links, in order to avoid creating multiple copies of the dynamic loader.

PR #699, opened 2025-08-06
Retag images as latest when built from the master branch

hjiawei:


images/Makefile
```diff
@@ -63,10 +63,16 @@ sub-calico-go-build-image-%:
 .PHONY: calico-go-build-cd
 calico-go-build-cd: calico-go-build-image var-require-one-of-CONFIRM-DRYRUN var-require-all-BRANCH_NAME
 	$(MAKE) BUILD_IMAGES=$(CALICO_GO_BUILD) retag-build-images-with-registries push-images-to-registries IMAGETAG=$(BRANCH_NAME) EXCLUDEARCH="$(EXCLUDEARCH)"
+ifeq ($(BRANCH_NAME),master)
```
MichalFupso, line 66: Would it be better to store the imagetag in variable TAG and set that variable based on the branch name. Then the make build command doesn't have to be repeated and can utilize the TAG variable

TAG=$(BRANCH_NAME)
ifeq ($(BRANCH_NAME),master)
  TAG=latest
endif

$(MAKE) BUILD_IMAGES=$(CALICO_GO_BUILD) retag-build-images-with-registries push-images-to-registries IMAGETAG="$(TAG)" EXCLUDEARCH="$(EXCLUDEARCH)"

MichalFupso (approved):

PR #707, opened 2025-09-02
add gcloud and docker-credential-gcr for using docker credHelper

radTuti:
This allow using docker credential helper for GCR/GAR.

There are 2 ways to use credential helper for GCR/GAR - gcloud CLI or docker-credential-gcr (standalone). This adds both libraries. This will require modifying the usage of go-build in the code to also mount the gcloud config.

```sh
docker run -t --entrypoint /bin/sh -v $(DOCKER_CONFIG):/root/.docker/config.json -v $(GCLOUD_CONFIG):/root/.config/gcloud go-build -c "crane cp SRC DST"
```

### DIFF
--- a/images/calico-go-build/Dockerfile
+++ b/images/calico-go-build/Dockerfile
@@ -6,6 +6,7 @@ FROM registry.access.redhat.com/ubi8/ubi:latest AS ubi
 
 ARG TARGETARCH
 
+ARG DOCKER_CREDHELPER_GCR_VERSION=v2.1.30
 ARG CONTAINERREGISTRY_VERSION=v0.20.6
 ARG CONTROLLER_TOOLS_VERSION=v0.18.0
 ARG GO_LINT_VERSION=v2.3.1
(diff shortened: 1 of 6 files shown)


images/calico-go-build/Dockerfile
```diff
@@ -36,6 +37,26 @@ RUN dnf upgrade -y && dnf install -y \
     xz \
     zip
 
+# Install Google Cloud SDK for GCR/GAR
+RUN cat <<EOM > /etc/yum.repos.d/google-cloud-sdk.repo
+[google-cloud-cli]
+name=Google Cloud CLI
+baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el9-x86_64
```
hjiawei, line 44: > The following sample command is for a Red Hat Enterprise Linux 9-compatible installation

It appears the instructions were adapted from [here](https://cloud.google.com/sdk/docs/install#rpm). However, our environment is still based on RHEL 8 and supports multiple architectures. We need a more generic repo configuration that is compatible with all supported platforms.

hjiawei:
Could you clarify why this is added to go-build and how this gcloud tool is used? Is it simpler to install on the semaphore runners (or maybe already installed)?

PR #721, opened 2025-10-07
deps: Bump go version to 1.24.8

tianfeng92:
Bump go version to 1.24.8.


.semaphore/semaphore.yml
```diff
@@ -48,7 +48,7 @@ blocks:
       env_vars:
         # The branch to test the current go-build against
         - name: CALICO_BRANCH
-          value: master
+          value: release-v3.31
```
tianfeng92, line 51: The Calico master branch now uses Go 1.25, which is incompatible with the 1.24 branch.

PR #817, opened 2026-04-27
[EV-6581] Add aarch64 cross-compilation sysroot to calico-rust-build

radixo:
Mirrors the calico-go-build cross-build setup: install clang/lld, extract an AlmaLinux 9 aarch64 sysroot under /usr/aarch64-linux-gnu, and bake cargo + cc-rs config so `cargo build --target aarch64-unknown-linux-gnu` works on amd64 hosts with no extra env.


.semaphore/promotions/calico-rust-build.yml
```diff
@@ -27,19 +27,6 @@ blocks:
...
-    dependencies:
-      - Publish calico/rust-build images
-    run:
-      when: "branch = 'master' AND change_in('/images/calico-rust-build/')"
-    task:
-      jobs:
-        - name: Linux multi-arch manifests
-          commands:
-            - if [ -z "${SEMAPHORE_GIT_PR_NUMBER}" ]; then make -C images push-calico-rust-build-manifests CONFIRM=true; fi
+            - if [ -z "${SEMAPHORE_GIT_PR_NUMBER}" ]; then make -C images calico-rust-build-cd CONFIRM=true; fi
```
hjiawei: Let’s keep the multi-arch builds in case people want to run native builds. The Rust builder image build doesn’t take long.

.semaphore/semaphore.yml
```diff
@@ -84,12 +84,9 @@ blocks:
...
             fi
       jobs:
-        - name: Build calico/rust-build image
+        - name: Build calico/rust-build amd64 image
           commands:
-            - make -C images calico-rust-build-image ARCH=$ARCH
-          matrix:
-            - env_var: ARCH
-              values: ["amd64", "arm64", "ppc64le", "s390x"]
+            - make -C images calico-rust-build-image
```
hjiawei: Same here. Let' keep the multi-arch image build and push.

images/calico-rust-build/Dockerfile
```diff
@@ -88,4 +155,16 @@ ENV RUSTUP_HOME=/usr/local/rustup
...
+# Note: cc-rs auto-adds --target= when CC=clang, so we only set --sysroot
+# here. -fuse-ld=lld routes link steps through LLD (multi-arch capable);
+# the system /usr/bin/ld is x86_64-only and rejects aarch64 emulations.
+ENV CC_aarch64_unknown_linux_gnu=clang \
+    CXX_aarch64_unknown_linux_gnu=clang++ \
+    AR_aarch64_unknown_linux_gnu=aarch64-unknown-linux-gnu-ar \
+    CFLAGS_aarch64_unknown_linux_gnu="--sysroot=/usr/aarch64-linux-gnu/sys-root" \
+    CXXFLAGS_aarch64_unknown_linux_gnu="--sysroot=/usr/aarch64-linux-gnu/sys-root" \
+    BINDGEN_EXTRA_CLANG_ARGS_aarch64_unknown_linux_gnu="--target=aarch64-linux-gnu --sysroot=/usr/aarch64-linux-gnu/sys-root -I/usr/aarch64-linux-gnu/sys-root/usr/include"
+
```
hjiawei: It's better to set these environment variables during the build rather than in this builder image. These changes are too specific to the arm64 architecture.

images/calico-rust-build/Dockerfile
```diff
@@ -4,15 +4,32 @@ ARG TARGETARCH
...
 
+ENV CC=clang
+ENV CXX=clang++
+
 RUN dnf upgrade -y && dnf install -y epel-release && \
     dnf install -y yq
 
 RUN dnf install -y \
+    binutils-aarch64-linux-gnu \
     bsdtar \
```
hjiawei, line 14: This is the only arm64-specific package in this generic install block, and nothing else here pulls it in transitively. On the arm64 variant of the image it's dead weight.

Suggest moving it into the cross-compile block below (around L80, next to `rustup target add aarch64-unknown-linux-gnu`) — that's where all the "this image cross-compiles for aarch64" logic should live, and it's already gated on `TARGETARCH=amd64`.

images/calico-rust-build/Dockerfile
```diff
@@ -4,15 +4,32 @@ ARG TARGETARCH
...
     wget
 
+# clang uses the verbatim target triple as the binutils prefix, so
+# `--target=aarch64-unknown-linux-gnu` looks for `aarch64-unknown-linux-gnu-ld`.
+# The EPEL package only ships the shorter `aarch64-linux-gnu-` prefix; bridge
+# the two with symlinks so clang picks the cross ld/as/ar instead of the host
+# /usr/bin/ld (which rejects aarch64 emulations).
+RUN set -eux; \
+    for tool in ld ld.bfd as ar nm objcopy objdump ranlib readelf strip; do \
+        ln -sfn "aarch64-linux-gnu-${tool}" "/usr/bin/aarch64-unknown-linux-gnu-${tool}"; \
```
hjiawei: Most of these symlinks aren't actually invoked at build time:

- `ld` / `ld.bfd`: bypassed — calico#12632's RUSTFLAGS sets `-fuse-ld=lld`, so clang uses lld and never searches for the GNU linker.
- `as`: bypassed — clang's integrated assembler handles assembly.
- `strip`: bypassed — rustc strips via `-Cstrip` using LLVM internals.
- `nm` / `objcopy` / `objdump` / `ranlib` / `readelf`: only invoked if a `build.rs` calls them directly, which ztunnel's deps don't do.
- `ar`: the only one cc-rs actually invokes (via `AR_<target>`). The **value** of that env var in 12632 can just be the EPEL name (`aarch64-linux-gnu-ar`) — see the matching comment on 12632. With that, the entire `RUN` block here can go.

If something downstream ever does call `aarch64-unknown-linux-gnu-<tool>` by long-triple name, we add the one targeted symlink at that point rather than carrying nine of them defensively.

hjiawei (commented):
Two suggestions — both about keeping the image neutral and pushing arch-specific config into the consumer Makefile (calico#12632). See inline.

hjiawei (approved):

PR #830, opened 2026-05-07
feat(tinygo): add calico/tinygo image for WebAssembly builds

electricjesus:
## Summary

- Adds a dedicated `calico/tinygo` image (mirrors `calico/rust-build`) packaging TinyGo 0.34.0 + Go 1.23.12 on AlmaLinux 9.
- amd64 only — wasm output is architecture-agnostic so a single host arch suffices for any consumer (Envoy V8 JITs for host arch at load time).
- First consumer: `gateway/coraza-wasm` (Calico Enterprise WAF), currently pinned to the upstream `tinygo/tinygo:0.34.0` image; this lets it switch to a Tigera-published toolchain image.
- CI builds the image and then compiles a small fixture against it that exercises the prebuilt-wasm-static-archive consumers `gateway/coraza-wasm` depends on, then asserts the produced wasm has no `env.cre2_*` imports — catching both wasm-ld link-time regressions and runtime missing-import regressions ("missing import: env.cre2_new") at toolchain-image build time rather than during a downstream WAF rebuild.

## Why TinyGo 0.34 (not the latest 0.41)

- `wasilibs/go-re2` v1.8.0+ removed TinyGo support because the underlying libre2 now requires Abseil with thread synchronization, which TinyGo's libc++ shim does not provide. Coraza's wasm path is therefore frozen on `go-re2` v1.6.0 + TinyGo 0.34 until upstream `coraza-proxy-wasm` migrates GC.
- `wasilibs/nottinygc` is archived/EOL since March 2024 and v0.7.1 is locked to TinyGo 0.34's wasi-libc ABI.
- Bake Go 1.23.12 directly (the highest patch in TinyGo 0.34's accepted host range; coraza-wasm's go.mod requires `>= 1.23.8`) instead of bundling a newer Go and forcing `GOTOOLCHAIN` switching at build time. The latter would require write access to `GOPATH/pkg/sumdb` at runtime, which downstream consumers running the container as `--user $(id -u):$(id -g)` cannot reliably provide.

When TinyGo relaxes its host-Go upper bound or wasilibs reverses course on TinyGo support, bump in lockstep.

## Test plan

- [x] Verified locally: `tinygo build -target=wasip1 -gc=precise ...` produces a valid 3.9MB wasip1 wasm with no `env.cre2_*` imports (cre2 resolves internally to the bundled `libcre2.a` static archive).
- [x] Verified locally: `gateway/coraza-wasm` full upstream package builds against this image and passes the existing CRS ftw baseline (3955/3955 tests, 0 skipped, 55 ignored, 17.3s).
- [ ] CI: `calico/tinygo image` block builds the image on amd64.
- [ ] CI: `Build and test calico/tinygo` job compiles the fixture against the just-built image and asserts no `env.cre2_*` imports.
- [ ] On master merge: `Publish calico/tinygo images` promotion publishes `calico/tinygo:<tag>` to the dev registry.

## Follow-ups (not in this PR)

- Consumer-side bump in `tigera/calico-private` (PR #11745, `gateway/coraza-wasm` Makefile TODO at line 9) to switch from `tinygo/tinygo:0.34.0` to `calico/tinygo:<tag>` once this image is published. Banked locally; will be pushed after merge.


images/calico-tinygo/versions.yaml
```diff
@@ -0,0 +1,10 @@
+tinygo:
+  version: 0.41.1
+  checksum:
+    sha256:
+      amd64: e156d1d93a376eef639a4143d13be07e8c463fb6cf2d7d447698ed4474d23e91
+golang:
```
hjiawei, line 18: Do we need both tinygo and official golang in one calico/tinygo image? How is tinygo used in the CI pipeline?

images/calico-tinygo/Dockerfile
```diff
@@ -0,0 +1,68 @@
+ARG TARGETARCH=${TARGETARCH}
+
+FROM almalinux:9 AS builder
+
+ARG TARGETARCH
+
+# calico/tinygo publishes a single host-arch image because TinyGo wasm output
+# is architecture-agnostic — the same .wasm bytes execute on amd64, arm64,
+# ppc64le and s390x runtimes (Envoy V8 JITs for the host arch at load time).
+# Building on arm64/ppc64le/s390x would just duplicate CI work without
```
hjiawei: I believe the main purpose of the multi-arch toolchain image (like linux/arm64 etc) is to allow developers on non-amd64 Linux systems to use native builder images. Since tinygo supports Linux amd64 and arm64, we can provide multi-arch images for them.

images/calico-tinygo/test/fixture/main.go
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// fixture exercises the calico/tinygo image against the failure class
+// where TinyGo's bundled wasi-libc and the prebuilt static archives shipped
+// inside wasilibs Go modules disagree on the libc/libc++ ABI. A trivial
+// smoke (`package main; func main(){}`) does NOT catch this — TinyGo's
+// stdlib provides everything the smoke needs.
+//
+// The regression only fires when wasm-ld has to resolve symbols out of
+// vendored `wasilibs/<lib>/wasm/*.a` archives. This fixture forces those
```
hjiawei: Where the *.a artifacts revoced from go-re2 in https://github.com/wasilibs/go-re2/pull/161 ~2 years ago (releasted in v1.8.0)? Do we still need this fixture validation test file?

images/calico-tinygo/versions.yaml
electricjesus: Yes — TinyGo internally invokes the host `go` toolchain to compile Go packages (parse, type-check, run `go mod` resolution) before lowering through LLVM IR to wasm, so the bundled `go` is not optional. Without it tinygo errors out at `go: command not found`.

In CI it's used by the `Build and test calico/tinygo` block in semaphore.yml: builds the image, then docker-runs the just-built tinygo against `images/calico-tinygo/test/fixture` to compile a wasm that exercises the same wasilibs static-archive consumers gateway/coraza-wasm depends on.

Added a clearer comment in commit `cc97d85` explaining the dependency (`Dockerfile:30`).

images/calico-tinygo/Dockerfile
electricjesus: Good catch — fixed in commit `cc97d85`. Now publishes amd64 + arm64 with a multi-arch manifest. Verified locally on an arm64 mac: `make calico-tinygo-image ARCH=arm64` builds clean, fixture compiles against the arm64 image and produces identical wasm bytes to the amd64 build (confirming the arch-neutrality of wasm output while still letting arm64 devs avoid qemu).

images/calico-tinygo/test/fixture/main.go
electricjesus: Correct. wasilibs/go-re2#161 removed the prebuilt `.a` archives in v1.8.0+, and the original framing of this fixture (catching wasm-ld undefined-symbol errors) doesn't apply once we've pinned to v1.6.0.

But the fixture is still load-bearing for a different failure mode: it asserts that the **TinyGo we ship in this image still links cleanly against go-re2 v1.6.0's bundled archive** and produces zero `env.cre2_*` imports. The negative path here isn't "someone bumps go-re2" — `gateway/coraza-wasm` won't accidentally do that since v1.7.x is the last TinyGo-compatible version. The negative path is "someone bumps the TinyGo version in this image to a release whose `wasi-libc` no longer matches what v1.6's static archives expect". That regression would surface here as either a wasm-ld link failure or as `env.cre2_*` imports leaking through, with the failure attributed to the *toolchain image* PR rather than to a downstream WAF rebuild.

In other words: post-v1.8 wasilibs design changes mean go-re2 itself is frozen for our use, so the fixture has effectively become a TinyGo-vs-v1.6.0-archive compat check. I'll update the comment header to say that explicitly — the current text is stale wrt the v1.8 change and probably caused this confusion. Thanks for the prompt!

hjiawei (approved):

PR #874, opened 2026-07-27
fix: update cron schedule to run at 4 AM UTC for version checks

dimitri-nicolo:


.github/workflows/check-version-updates.yml
```diff
@@ -2,7 +2,7 @@ name: Check for Go and Kubernetes version updates
 
 on:
   schedule:
-    - cron: '0 1 * * *'
+    - cron: '0 4 * * *'
```
MichalFupso, line 5:
```suggestion
    - cron: '0 4 * * *'
      timezone: 'America/Vancouver'
```

dimitri-nicolo: @MichalFupso Shouldn't it be 12 AM America/Vancouver timezone?

MichalFupso (approved):

PR #907, opened 2026-08-26
Add CI utility scripts to the go build image from a versioned container

RoryDoherty:
Copies the scripts in from a versioned image instead of trying to keep the scripts updated in two locations at once


images/calico-go-build/Dockerfile
```diff
@@ -306,9 +311,14 @@ COPY bin/semvalidator-${TARGETARCH} /usr/local/bin/semvalidator
 # Install Argo CI helper scripts so this image can serve as the base for Argo
 # workflows, which have no pre-checkout step. Semaphore runners do their own
 # checkout and ignore these.
-COPY scripts/checkout.sh /usr/local/bin/checkout
-COPY scripts/setup-github-ssh.sh /usr/local/bin/setup-github-ssh
-COPY scripts/create-local-secret.sh /usr/local/bin/create-local-secret
+COPY --from=ci-scripts / /usr/local/bin/
```
hjiawei: You can copy from /from-script-name to /usr/local/bin/to-script-name so no need to run ln later.

images/calico-go-build/Dockerfile
```diff
@@ -50,6 +54,7 @@ RUN set -eux; \
         pcre-devel \
         pigz \
         pkg-config \
+        sudo \
```
hjiawei, line 57: Why do we need sudo now?

images/calico-go-build/Dockerfile
RoryDoherty: the ln is to allow the script to be called with either name
The scripts that were removed are using kebab-case for the names whereas ArgoCI uses camelCase for the script names
I added the ln so that it doesn't break any existing CI that might be using the kebab-case method calls

images/calico-go-build/Dockerfile
RoryDoherty: The checkout script uses sudo to chmod on the checked out directory to give broader write permissions
https://github.com/tigera/cc-utils/blob/tsla-11651-pre/argoci-images/common-scripts/checkout#L34

we do support no-sudo as an option so if you would prefer I can remove this

images/calico-go-build/Dockerfile
hjiawei: OK. It is better to do an explicit copy from the source to the target (like the lines you removed), rather than all files under `/` to `/usr/local/bin`. We don't want files or folders to get overwritten without realizing it. Being explicit will also make the build fail loudly if the source file doesn't exist.

RoryDoherty: Updated now so that each script is copied over explicitly

hjiawei (approved):

images/calico-go-build/Dockerfile
```diff
@@ -1,5 +1,9 @@
 ARG TARGETARCH=${TARGETARCH}
 
+# Bump to pick up a change to the Argo CI helper scripts copied in near the end.
+ARG CI_SCRIPTS_VERSION=v1
+FROM calico/ci-scripts:${CI_SCRIPTS_VERSION} AS ci-scripts
```
hjiawei, line 5: Failed for ppc64le and s390x:

> failed to solve: calico/ci-scripts:v1: failed to resolve source metadata for docker.io/calico/ci-scripts:v1: no match for platform in manifest: not found

PR #908, opened 2026-09-02
Add scratch-utils CI image, ci-base VM builder, and go-build preload disk

Brian-McM:
## Summary

Adds the CI VM tooling the ArgoCI kind-rig needs, as three pieces that share one
source of truth for versions:

- **`scratch-utils`** — one static binary (`createvm` / `deletevm` / `secret` /
  `runonvm`) that drives the GCE VM lifecycle over the GCP APIs directly, with no
  gcloud and no bash, so it runs from a distroless image. Published as
  `calico/scratch-utils`.
- **`vm-images/ci-base/`** — builds the `ci-base` GCE image: docker, go, kind, kubectl and
  gh baked in, plus the heavy CI container images pre-pulled, so a CI VM boots
  ready and does zero per-run installs.
- **`preload-disks/ci-cache/`** — builds a GKE secondary-boot-disk image with
  `calico/go-build` preloaded, so build pods start with it already on the node
  instead of pulling several hundred MB per run.

It's CI VM tooling that belongs alongside `go-build`, so it's folded into this
module rather than a separate one.

## Layout

Both builder trees are plural with the artifact as a subfolder, so a second one
lands as a sibling rather than a rewrite:

```
vm-images/ci-base/        # the GCE image family it produces
preload-disks/ci-cache/   # what it caches
```

Each subfolder is named for the axis a sibling would vary on. No shared driver is
factored out yet — with one type in each tree that would be guessing at what the
second needs, and the two trees share nothing anyway (one drives gcloud, the other
drives daisy through a fetched Go tool).

## Everything derives from `images/calico-go-build/versions.yaml`

This is the part worth reviewing closely. A job that runs on one of these VMs
would otherwise have run inside `calico/go-build`, so it has to see the same Go —
which means the VM's toolchain must track that file, not this module's `go.mod`.

`vm-images/ci-base/build-image.sh` resolves the Go version, its checksum and the
`calico/go-build` tag from it (via the existing `hack/generate-version-tag-name.sh`,
the same helper the release tagging uses) and injects them into `provision.sh`,
which runs as the builder VM's startup-script where it cannot read the repo.
`preload-disks/ci-cache` resolves the same tag for the image it preloads. Both **require**
those values rather than defaulting them: silently baking a stale Go is the exact
drift this indirection exists to prevent.

Tools with no home in that file — kind, gh, the kind node image — are pinned in
`vm-images/ci-base/versions.yaml`. Nothing floats on `latest` or `stable.txt`, so two
builds of one commit produce the same image.

## Image naming

Images are named off the go-build release tag, so a VM image and the toolchain it
was built against are matchable by eye:

| | |
|---|---|
| go-build image | `calico/go-build:1.27.1-llvm21.1.8-k8s1.37.0` |
| ci-base VM image | `ci-base-1-27-1-llvm21-1-8-k8s1-37-0` |
| preload disk | `cic-1-27-1-llvm21-1-8-k8s1-37-0` |

`hack/generate-image-name.sh` builds these. Two constraints shape it: GCE names
are RFC1035 so the dots become hyphens (the exact tag is kept as a `go-build-tag`
label), and **GKE caps a secondary boot disk name at 39 characters** — enforced
when a node pool *attaches* the image, long after it built cleanly — which is why
the preload prefix is terse and why the generator takes `-m` to fail at build time
instead.

On a tag build the version comes from `$SEMAPHORE_GIT_TAG_NAME`, matching what
`calico-go-build-cd` already does; that is the only place the re-release suffix
lives (`-1`, `-2`), which `generate-version-tag-name.sh` cannot know about.
Otherwise it uses the branch.

## CI

- Build check for the `calico/scratch-utils` image on every PR.
- New **`go test`** block, run in the `golang` image at the version `versions.yaml`
  pins — so tests see the same Go the go-build image ships and the agent needs
  none of its own. This repo ran no Go tests before.
- `promotions/calico-scratch-utils.yml` — change-gated publish, mirroring the
  other image promotions.
- `promotions/ci-base-vm-image.yml` and `promotions/preload-disk-image.yml` —
  **chained off the go-build publish**, auto-promoted on a release tag, so both
  images rebuild once the go-build image they bake or cache actually exists.
  Gated on the tag rather than merely on that pipeline passing, since it also runs
  for master, where a rebuild would spend a builder VM per merge for a throwaway
  branch image. Both remain promotable by hand for a one-off rebuild.

## New dependencies

`google.golang.org/api` (compute) and `golang.org/x/crypto` (ssh). Both are for
`scratch-utils`; nothing else in the module imports them.

Deliberately **not** added: `gke-disk-image-builder`. It has a clean library API,
but its `go.mod` still declares the old `GoogleCloudPlatform/ai-on-gke` path while
the code lives at `ai-on-gke/tools`, so `go get` at the real location fails on a
path mismatch and the declared path resolves only to an abandoned 2023 snapshot.
`preload-disks/ci-cache` fetches a pinned commit instead, which also keeps `compute-daisy`
and the `cloud.google.com/go` stack — about 17 modules — out of this module's
dependency graph.

## Testing

- 21 unit tests across 6 files: `gofmt`, `go vet` and `go test` clean on Go 1.27.1.
  Coverage is the pure logic plus the security-relevant path — `untar`'s guard
  against tar entries escaping the destination, which `GetDir` applies to whatever
  a remote sends. That test was mutation-checked: it fails when the guard is
  removed. `shellQuote` is verified by round-tripping hostile values through a real
  bash, including inside the `export NAME=...` line the env file is built from.
- Both images were **actually built and used**: `ci-base-1-27-1-llvm21-1-8-k8s1-37-0`
  booted real CI VMs, and the preload disk builder was run end-to-end with only its
  defaults. The 1.27.1 rebuilds follow this merge.
- `scratch-utils:v0.3` built, pushed and pulled back to confirm the binary
  dispatches and errors cleanly without credentials.

## Notes for review

- **Subcommand, not command.** The image ENTRYPOINT is the binary, so a pod passes
  `args: [createvm]` (or `command: [scratch-utils, createvm]`). A bare
  `command: [createvm]` overrides the entrypoint and fails to exec.
- **`max-run-duration` is eventual, not guaranteed.** It fires on time, but it is
  itself an operation: observed in us-central1-a, an insert stayed RUNNING for over
  90 minutes and the deadline's own DELETE then sat PENDING behind it. Hence the
  per-zone timeout in the zone loop rather than relying on the backstop.
- **`GOOGLE_VM_IMAGE`** pins an exact VM image; unset, the `ci-base` family gives
  the newest. The pin is what lets the image be rolled without rebuilding the
  binary.
- The `kindest/node` tag is still hand-maintained in `vm-images/ci-base/versions.yaml` — it
  comes from calico's `lib/kind`, which this repo has no view of.

---
*Authored with AI assistance (Claude Code).*


vm-images/ci-base/README.md
```diff
@@ -0,0 +1,137 @@
...
+| go-build image | `calico/go-build:1.27.0-llvm21.1.8-k8s1.37.0` |
+| ci-base VM image | `ci-base-1-27-0-llvm21-1-8-k8s1-37-0` |
+
+`hack/generate-image-name.sh` builds the name. Dots become hyphens because GCE
+resource names are RFC1035 (lowercase, digits, hyphens, 63 max) — the exact,
+un-mangled tag is also attached as a `go-build-tag` label, so you can still look
+an image up by it:
+
+```bash
+gcloud compute images list --filter="labels.go-build-tag=1_27_0-llvm21_1_8-k8s1_37_0"
```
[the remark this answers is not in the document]
Brian-McM: Not changing this one — the constraint cited applies to label **keys**, not label
**values**. Label values may start with a digit and may contain underscores; it is
keys that must begin with a lowercase letter.

`build-image.sh` sets `go-build-tag=<tag with dots -> underscores>`, and that value
is live on a real image today:

```
$ gcloud compute images describe ci-base-1-27-1-llvm21-1-8-k8s1-37-0 \
    --project=unique-caldron-775 --format='value(labels)'
go-build-tag=1_27_1-llvm21_1_8-k8s1_37_0
```

So the example was accurate. Normalising it to a `v` prefix and hyphens would in
fact break the thing the label exists for: the image *name* already has dots
rewritten to hyphens for RFC1035, and the label is the one place the exact
go-build tag survives for lookup. Rewriting the separators would make it
ambiguous with the name again.

Resolving as won't-fix. The file itself has since been deleted (the READMEs were
dropped in 3d1b7ea), so the example is gone either way — noting the reasoning here
in case the label format gets documented again later.

scratch-utils/util/shell.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// ShellQuote single-quotes s for a POSIX shell: an embedded apostrophe is closed,
+// backslash-escaped and reopened, so any value survives verbatim. The literal
+// escape is only in the code below -- gofmt rewrites a doubled apostrophe in a
+// comment into a curly quote, so it cannot be spelled here.
+//
```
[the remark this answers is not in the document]
Brian-McM: gofmt does modify comment typography — but only in **doc comments**. Since Go 1.19
it reformats them and applies godoc's old quoting convention, which turns a doubled
apostrophe into a right curly quote (U+201D). Verified on go1.27.0:

```go
// DocComment turns ' into '\'' in a shell.
func DocComment() {}

func Inner() {
	// A plain inner comment: ' into '\'' here.
}
```

After `gofmt -w`, only the doc comment changes:

```
-// DocComment turns ' into '\'' in a shell.
+// DocComment turns ' into '\” in a shell.
```

The inner comment is untouched. This is not hypothetical — it is how the original
`'\”` in `shellQuote`'s doc comment got there, and it silently reverted an earlier
attempt to spell the escape correctly.

That said, the comment deserved the flag: it said "a comment" when it is
specifically a doc comment, which made the constraint sound broader and stranger
than it is. Reworded in 4387ea8 to name doc comments and to note that a comment
inside a function body would be left alone.

hjiawei (commented):
Review of the scratch-utils / ci-base / preload-disk changes. Findings are inline. The one at `.semaphore/promotions/calico-go-build.yml:41` is blocking: as written, neither new pipeline can be reached.

.semaphore/promotions/calico-go-build.yml
```diff
@@ -28,6 +28,24 @@ global_job_config:
...
+# published -- chained here rather than off the root pipeline so they cannot start
+# before the image they consume exists.
+#
+# Gated on the release tag, not merely on this pipeline passing: it also runs for
+# master, where a rebuild would spend a builder VM per merge and, with no tag set,
+# would only produce the throwaway ci-base-master / cic-master branch images.
+# Both stay promotable by hand from the Semaphore UI for a one-off rebuild.
+promotions:
+  - name: Build ci-base VM image
+    pipeline_file: promotions/ci-base-vm-image.yml
```
hjiawei: Semaphore resolves a relative `pipeline_file` from the directory of the *current* pipeline, not from `.semaphore/`: ["If `pipeline_file` is a relative path, Semaphore will search for the file inside the directory of the current pipeline."](https://docs.semaphore.io/reference/pipeline-yaml)

This file is already inside `promotions/`, so both paths here resolve to `.semaphore/promotions/promotions/...`, which doesn't exist. Neither pipeline can be reached — not by auto-promote on a release tag, and not by the manual promotion the comment above offers as the one-off escape hatch. `.semaphore/semaphore.yml` writes `promotions/calico-base.yml` correctly because it sits in `.semaphore/`.

Use `pipeline_file: ci-base-vm-image.yml` and `preload-disk-image.yml`, or absolute `/.semaphore/promotions/...`.

.semaphore/promotions/calico-go-build.yml
```diff
@@ -28,6 +28,24 @@ global_job_config:
...
+    auto_promote:
+      when: "result = 'passed' AND tag =~ '^1\\.\\d+\\.\\d+-llvm\\d+\\.\\d\\.\\d-k8s1\\.\\d+\\.\\d+'"
```
hjiawei: `llvm\d+\.\d\.\d` only matches a single-digit LLVM minor and patch. On a bump to llvm 21.1.10 the regex consumes `llvm21.1.1` and then needs `-k8s1.` where the tag has `0-k8s1.`, so it stops matching.

Nothing errors when that happens: `auto_promote` just never fires, and the identical `run.when` gate in `ci-base-vm-image.yml:48` / `preload-disk-image.yml:54` also goes false, so even a manual promotion gives a green pipeline with the block skipped. The release ships with no ci-base image and no preload disk.

`llvm\d+(\.\d+){2}` fixes it. Four copies of this pattern now exist — worth a single anchor they all reference.

scratch-utils/gce/vm.go
```diff
@@ -0,0 +1,230 @@
...
+func (c *Client) FindZone(ctx context.Context, name string) (string, error) {
+	// The value must be quoted -- an unquoted name with a - or . is a syntax error,
+	// not a non-match -- and AggregatedList pages even for a single match.
+	call := c.svc.Instances.AggregatedList(c.project).Filter(fmt.Sprintf("name=%q", name))
+	for {
+		agg, err := call.Context(ctx).Do()
+		if err != nil {
+			return "", err
+		}
+		for scope, list := range agg.Items {
```
hjiawei, line 202: Go randomizes map iteration order, so returning on the first non-empty scope picks an arbitrary zone when the same VM name exists in two.

That isn't hypothetical here: `PerZoneTimeout` (3m) can leave an insert still RUNNING in an earlier zone whose best-effort DELETE is also still PENDING — the 85-minute us-central1-a case the comment at :55 describes — while `Create` succeeds in the next zone. Instance names are zone-scoped, so GCE happily holds both. `deletevm` with `ZONE` unset then deletes one at a coin flip and leaks the other, and `runonvm` with `ZONE` unset can SSH into the abandoned VM and run the job on a machine `createvm` never reported.

Collect every matching zone, then either fail loudly on >1 or act on all of them.

scratch-utils/gce/ssh.go
```diff
@@ -0,0 +1,358 @@
...
+	stdout, err := sess.StdoutPipe()
+	if err != nil {
+		return err
+	}
+	// Tar the contents (not the dir itself) so they land directly under localDir.
+	cmd := fmt.Sprintf("cd %s 2>/dev/null && tar czf - . || true", util.ShellQuote(remoteDir))
+	if err := sess.Start(cmd); err != nil {
+		return err
+	}
+	if err := untar(stdout, localDir); err != nil {
```
hjiawei, line 312: This deadlocks. `StdoutPipe`'s contract is that the reader must keep draining or the remote command blocks; on the `untar` error path the pipe is abandoned with the remote `tar czf - .` still writing. The SSH window fills, the remote tar blocks on write and never exits, and `Wait` never returns.

Reachable from the `escapes dest` guard, ENOSPC locally, or a permission error creating a target file. Because `GetDir` runs from `runonvm`'s deferred `--get` loop, the hang happens *after* the user's script finished and reported its exit code — it looks like the job wedged at 100%.

`io.Copy(io.Discard, stdout)` or `sess.Close()` before `Wait` on the error path.

scratch-utils/gce/ssh.go
```diff
@@ -0,0 +1,358 @@
...
+		Auth:            []ssh.AuthMethod{ssh.PublicKeys(signer)},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(), //nolint:gosec // ephemeral CI VM, no key to pin
+		Timeout:         10 * time.Second,
+	}
+	addr := net.JoinHostPort(ip, "22")
+
+	deadline := time.Now().Add(3 * time.Minute)
+	var lastErr error
+	for time.Now().Before(deadline) {
+		client, err := ssh.Dial("tcp", addr, cfg)
```
hjiawei: `ssh.ClientConfig.Timeout` bounds only the TCP connect — `ssh.Dial` passes it to `net.DialTimeout` and then runs `NewClientConn` with no deadline on the conn.

The state this loop exists to ride out is exactly the one that hangs: a freshly-booted VM whose sshd has bound the socket but isn't ready to send its version string accepts the TCP connection and then stalls. `ssh.Dial` never returns, so neither the `deadline` check nor the `ctx.Done()` select below is ever reached, and `runonvm`'s 10-minute `setupCtx` doesn't fire either. The step hangs until CI kills it.

Dial with `net.DialTimeout`, `conn.SetDeadline`, then `ssh.NewClientConn` — or race each attempt against `ctx` in a goroutine.

scratch-utils/gce/ssh.go
```diff
@@ -0,0 +1,358 @@
...
+	signer, authorized, err := ephemeralKey()
+	if err != nil {
+		return nil, err
+	}
+	ip, err := c.injectKeyAndGetIP(ctx, zone, name, user, authorized)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg := &ssh.ClientConfig{
```
hjiawei, line 51: No keepalive is configured, and x/crypto/ssh sends none on its own. GCP VPC connection tracking drops an idle established TCP flow after 10 minutes, and that isn't tunable.

This connection carries an unbounded CI job, so any quiet stretch over 10 minutes — a long `go build`, `kind create cluster` waiting on the control plane, a big image pull — tears it down. `conn.Run` returns a transport error and the deferred `--get` sessions fail too, so the job dies with no logs pulled back.

A goroutine sending `client.SendRequest("keepalive@openssh.com", true, nil)` every ~60s for the life of the session covers it.

vm-images/ci-base/provision.sh
```diff
@@ -0,0 +1,77 @@
...
+EOF
+
+# --- pre-pull the heavy CI images so jobs skip the pull ---------------------
+# Tags are injected by build-image.sh, so they cannot drift.
+systemctl start docker
+PREPULL_IMAGES=("$GO_BUILD_IMAGE" "registry:2")
+# Unquoted on purpose: KIND_NODE_IMAGES is a space-separated list.
+for img in $KIND_NODE_IMAGES; do PREPULL_IMAGES+=("$img"); done
+for img in "${PREPULL_IMAGES[@]}"; do
+  retry docker pull "$img" || echo "warn: could not pre-pull $img"
```
hjiawei: `|| echo` neutralizes `set -e`, so provision.sh goes on to touch the readiness marker after all 8 retries for a kind node image have failed — a registry blip, a rate limit, or a bad digest in versions.yaml.

`build-image.sh:109` then verifies only `$GO_BUILD_IMAGE`, never the node images or `registry:2`, so the image snapshots and publishes to the `ci-base` family looking fine. Every VM booted from it pulls ~1.5GB per kind cluster at run time — the exact cost this pre-pull exists to remove — and it only ever shows up as jobs being slower.

That verify block was added for this class of bug; the node images came later and never got added to the check. Worth asserting every entry of `PREPULL_IMAGES`.

vm-images/ci-base/build-image.sh
```diff
@@ -0,0 +1,132 @@
...
+  printf 'export KIND_NODE_IMAGES=%q\n' "$KIND_NODE_IMAGES"
+  printf 'export GH_VERSION=%q\n' "$GH_VERSION"
+  tail -n +2 "$HERE/provision.sh" # its shebang is replaced by the one above
+} >"$STARTUP"
+
+cleanup() {
+  rm -f "$STARTUP"
+  gcloud --quiet compute instances delete "$BUILDER" --project="$PROJECT" --zone="$ZONE" 2>/dev/null || true
+}
+trap cleanup EXIT
```
hjiawei: `trap cleanup EXIT` doesn't fire on SIGTERM — bash only runs the EXIT trap on TERM if TERM is trapped too.

The provision poll below is 120 iterations of a 10s-ConnectTimeout ssh plus a 10s sleep, so up to ~40 minutes against the `execution_time_limit: minutes: 30` in ci-base-vm-image.yml. Slow provisioning (apt mirror, docker pulls) therefore ends in Semaphore SIGTERMing the job with `cleanup` never running. Unlike the VMs `createvm` makes, this builder is a plain `gcloud compute instances create` with no `--max-run-duration`, so an e2-standard-8 runs until someone notices.

`trap cleanup EXIT INT TERM`. `preload-disks/ci-cache/build-preload-disk.sh:52` has the same gap, and additionally abandons the daisy builder VM and disk that `go run ./cli` created.

scratch-utils/subcmd/createvm/createvm.go
```diff
@@ -0,0 +1,128 @@
...
+		return fmt.Errorf("VM_NAME must be set")
+	}
+	project := envOr("GCP_VM_PROJECT", "unique-caldron-775")
+	zoneOut := envOr("ZONE_OUT", "/tmp/vm-zone")
+	// Point ADC at the compute SA (mounted key file, or materialized from its env var).
+	if err := util.SetupComputeADC(); err != nil {
+		return err
+	}
+
+	maxRun, err := time.ParseDuration(envOr("GOOGLE_VM_MAX_RUN_DURATION", "90m"))
```
hjiawei: `time.ParseDuration` accepts `0s`, `-5m` and `-1h` without error. `instanceSpec` then sees `cfg.MaxRun > 0` as false and omits the whole `Scheduling` block — no ProvisioningModel, no InstanceTerminationAction, no MaxRunDuration. The VM creates fine and looks normal, but has no reclaim deadline at all.

That matters because everything else leans on this backstop: `deletevm` returns 0 on any failure with "(leaving to max-run-duration)", and `vm.go:112` logs "max-run-duration is the backstop" after a failed cleanup. With it gone, an n2-standard-16 runs until someone notices.

`parseDiskGB` below already rejects `n <= 0` for exactly this reason — same guard here.

scratch-utils/gce/ssh.go
```diff
@@ -0,0 +1,358 @@
...
+		op, err = c.svc.Instances.SetMetadata(c.project, zone, name, md).Context(ctx).Do()
+		return err
+	}); err != nil {
+		return "", err
+	}
+	if err := c.waitZoneOp(ctx, zone, op.Name); err != nil {
+		return "", fmt.Errorf("set-metadata op on %s: %w", name, err)
+	}
+
+	ip := externalIP(inst)
```
hjiawei: `inst` was fetched before `SetMetadata` and its zone operation, which can take tens of seconds, so this reads a stale snapshot. Since `createvm` deliberately doesn't wait for readiness (`DialSSH`'s retry is the readiness check), `runonvm` can `Get` the instance right after the insert op reaches DONE, when `natIP` may not be populated yet.

The result is a terminal `instance %s has no external IP` with no retry path, failing the step even though the IP shows up seconds later and `DialSSH` would have handled the rest.

Read the IP from a `Get` issued after the metadata op, or fold it into the existing `retry`.

scratch-utils/subcmd/runonvm/runonvm.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+
+	// A file, not a command string.
+	remoteScript := path.Join("/tmp", path.Base(script))
+	if err := conn.PutFile(script, remoteScript, 0o755); err != nil {
+		fmt.Fprintf(os.Stderr, "runonvm: upload script: %v\n", err)
+		return 1
+	}
+	runCmd := "bash " + remoteScript
```
hjiawei: These two are the only places the remote paths go into a shell command unquoted — `PutData` runs the same value through `util.ShellQuote` one call earlier (ssh.go:187).

`remoteScript` is `path.Join("/tmp", path.Base(script))` from caller-supplied `--script` or Argo's trailing arg. `--script '/work/my step.sh'` uploads to `/tmp/my step.sh` correctly (quoted) and then runs `bash /tmp/my step.sh`, which the remote shell splits in two: "No such file or directory" after every upload already succeeded, pointing at the wrong thing.

`util.ShellQuote` on both.

cmd/Makefile
```diff
@@ -17,6 +17,14 @@ bin/semvalidator-$(ARCH): semvalidator/main.go
 	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) \
 		go build -o bin/semvalidator-$(ARCH) -v -buildvcs=false -ldflags "-s -w" semvalidator/main.go
 
+# Unlike the single-file binaries above, this one spans ../scratch-utils/, so list
+# every .go file it compiles -- otherwise an incremental build ships a stale binary.
+SCRATCH_UTILS_SRCS = $(shell find scratch-utils ../scratch-utils -name '*.go')
```
hjiawei: Two things here:

`=` is recursive, so this `find` re-runs every time make expands the prerequisite list. `:=` runs it once.

Bigger: `build:` now depends on `bin/scratch-utils-$(ARCH)`, and `build` is a prerequisite of `calico-go-build-image` (images/Makefile:67) and `calico-binfmt-image` (:100). Every arch of the go-build image matrix now downloads and compiles `google.golang.org/api`, gRPC and the OTel stack — roughly 20 modules those images never contain — before it can start its own Dockerfile.

Give `bin/scratch-utils-$(ARCH)` its own target that only `calico-scratch-utils-image` depends on.

vm-images/ci-base/build-image.sh
```diff
@@ -0,0 +1,132 @@
...
+fi
+
+log "stopping builder for a consistent disk"
+gcloud --quiet compute instances stop "$BUILDER" --project="$PROJECT" --zone="$ZONE"
+
+# The name has dots rewritten for RFC1035; the label keeps the exact tag.
+log "creating image $IMAGE in family $FAMILY"
+gcloud compute images create "$IMAGE" --project="$PROJECT" \
+  --source-disk="$BUILDER" --source-disk-zone="$ZONE" --family="$FAMILY" \
+  --labels="go-build-tag=$(echo "$GO_BUILD_IMAGE" | sed 's|.*:||; s|\.|_|g')"
```
hjiawei: The comment above says the label keeps the exact tag, and the PR description repeats it, but the `sed` rewrites every dot: the stored value is `1_27_1-llvm21_1_8-k8s1_37_0`.

So `gcloud compute images list --filter="labels.go-build-tag=1.27.1-llvm21.1.8-k8s1.37.0"` returns nothing, which is the label's only stated purpose — the image *name* already carries the same munged form.

GCE label values allow dashes, so `s|\.|-|g` would at least match the name. Either way the comment should say what's actually stored.

scratch-utils/subcmd/createvm/createvm.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("GOOGLE_VM_DISK_SIZE %q: want e.g. 200GB: %w", s, err)
+	}
+	if n <= 0 {
+		return 0, fmt.Errorf("GOOGLE_VM_DISK_SIZE %q: must be positive", s)
+	}
+	return n, nil
+}
+
+func envOr(key, def string) string {
```
hjiawei: This same eight-line `envOr` is also in `deletevm.go:67` and `runonvm.go:225`, and all three files already import `scratch-utils/util` — whose doc comment says it's the home for exactly this. The empty-vs-unset semantics here are the distinction `LocalSecret` and `--put-env` care about, so a change to it now has to land in three places or the subcommands quietly diverge.

Separately, `createvm_test.go:51` hand-rolls `containsStr`; every other test file in the PR just uses `strings.Contains`.

scratch-utils/gce/vm.go
```diff
@@ -0,0 +1,230 @@
...
+func New(ctx context.Context, project string) (*Client, error) {
+	svc, err := compute.NewService(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("compute client: %w", err)
+	}
+	return &Client{svc: svc, project: project}, nil
+}
+
+// PerZoneTimeout bounds ONE zone's attempt, so a degraded zone can be skipped. A
+// single deadline across the loop let one slow zone consume all of it and blamed
```
hjiawei: This comment (and a few others) explains what the code used to do and why the old bug is gone, rather than the present state — that belongs in the commit message.

Same shape at `deletevm.go:32` ("pointing ADC at a nonexistent /secrets path made every delete fail auth..."), `createvm.go:63` ("a fixed cap silently starved the later zones..."), `build-image.sh:106` ("This used to end in `|| true`..."), and `util/secret.go:61` + `util/secret_test.go:71` ("The key used to go to a fixed /tmp path").

The forward-looking half of each is the part worth keeping — here, that the timeout is per-zone so a degraded zone can be skipped, and that 3 minutes is chosen because the failure mode is slowness.

.semaphore/promotions/calico-go-build.yml
Brian-McM: Not that this isn't true, but I copied this verbatim from our other pipelines so we should probably fix those there too.

.semaphore/promotions/calico-go-build.yml
Brian-McM: Fixed.

.semaphore/promotions/calico-go-build.yml
Brian-McM: I just fixed this for all instances of the regex in this PR

scratch-utils/gce/ssh.go
Brian-McM: Added a keep alive.

scratch-utils/gce/ssh.go
Brian-McM: Fixed.

scratch-utils/subcmd/createvm/createvm.go
Brian-McM: Fixed.

scratch-utils/gce/ssh.go
Brian-McM: Fixed.

scratch-utils/subcmd/createvm/createvm.go
Brian-McM: simplified.

vm-images/ci-base/build-image.sh
Brian-McM: So I did add the extra terms to the trap (just in case) but it seems like in Bash (no sh) this should fire on SIGTERM.

Besides that, I also added a max-run-duration on the build-image.sh script for creating the build vm so that GCP will reap any leaked build images itself.

scratch-utils/subcmd/runonvm/runonvm.go
Brian-McM: Fixed.

cmd/Makefile
Brian-McM: Fixed.

scratch-utils/gce/vm.go
Brian-McM: I trimmed all the comments.

vm-images/ci-base/provision.sh
Brian-McM: Done, this is a fatal error now.

scratch-utils/gce/vm.go
```diff
@@ -0,0 +1,309 @@
...
+type zoneInstance struct {
+	zone   string
+	status string
+}
+
+// goingAway reports a status from which the instance will not become usable --
+// it is shutting down, already stopped, or on its way out.
+func goingAway(status string) bool {
+	switch status {
+	case "STOPPING", "STOPPED", "SUSPENDING", "SUSPENDED", "TERMINATED":
```
hjiawei, line 227: Follow-up, not blocking merge.

The SDK's `Instance.Status` enum has two more values that belong here: **`DEPROVISIONING`** ("halted and performing tear down tasks") and **`PENDING_STOP`** ("gracefully shutting down"). `DEPROVISIONING` is what the orphan reports while its queued DELETE finally executes — the scenario this commit exists for. Treated as live, one `RUNNING` + one `DEPROVISIONING` trips the "exists in more than one zone" error: `runonvm` exits 1 with a good VM available, and `deletevm` leaves both to max-run-duration.

```go
case "STOPPING", "STOPPED", "SUSPENDING", "SUSPENDED", "TERMINATED", "DEPROVISIONING", "PENDING_STOP":
```

(`google.golang.org/api@v0.296.0/compute/v1/compute-gen.go`, the `Instance.Status` doc comment.)

hjiawei (approved):

PR #911, opened 2026-09-02
Auto-update Go/Kubernetes versions on the go1.27 branches

dimitri-nicolo:
Add  `go1.27` to the nightly check-version-updates.

### DIFF
--- a/.github/workflows/check-version-updates.yml
+++ b/.github/workflows/check-version-updates.yml
@@ -17,6 +17,8 @@ jobs:
       matrix:
         branch:
           - master
+          - go1.27
+          - go1.27-k8s1.36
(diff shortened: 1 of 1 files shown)


.github/workflows/check-version-updates.yml
```diff
@@ -17,6 +17,8 @@ jobs:
       matrix:
         branch:
           - master
+          - go1.27
+          - go1.27-k8s1.36
```
dimitri-nicolo: Removing `go1.27-k8s1.36`, as it may not be in use

hjiawei (approved):

MichalFupso (approved):

PR #917, opened 2026-09-16
Vendor gke-disk-image-builder and move it off Debian 11

Brian-McM:
The preload-disk promotion has been failing since #916 unblocked it. The failure
is in the builder VM, not our code.

## The break

Daisy tripped its own failure match on `startup-script-url exit status 1`. The
builder VM's serial log:

```
E: Failed to fetch https://deb.debian.org/debian-security/pool/updates/main/c/
containerd/containerd_1.4.13~ds1-1~deb11u6_amd64.deb  404  Not Found
Failed to start containerd.service: Unit containerd.service not found.
sudo: ctr: command not found
```

Upstream hardcodes `debian-11-bullseye-v20230912` — a September 2023 image — as the
boot disk for the VM that unpacks the container images. Three years on:

- that image is `DEPRECATED` in `debian-cloud` (it still boots, which is why we got
  this far),
- `bullseye-security`'s `Release` file still returns 200, so `apt update` succeeds
  and builds an index,
- but the `.deb` that index names is **404** — the pool now carries only `deb13`
  builds.

So containerd never installs, `ctr` doesn't exist, and the script exits before
pulling anything. The same pinned ref built `cic-1-27-1-llvm21-1-8-k8s1-37-0` fine
on Sept 9, so this broke on Debian's side within the last week.

## Why vendor rather than patch in passing

`462a06c` is the **tip** of upstream `main`, not a pin behind it. There is nothing
to bump to, so the fix is ours either way — and a `sed` against a freshly fetched
tree would have held only until upstream touched that line.

It also can't be an ordinary Go dependency: its `go.mod` declares the
`GoogleCloudPlatform/ai-on-gke` path, and that directory was deleted from that
repo, so `go get` reaches only an abandoned 2023 snapshot.

## What changed

- Boot image is now the **`debian-12` family**, not a dated image, so it tracks the
  newest member and cannot rot the same way. Bookworm rather than trixie because
  `script/startup.sh` was written against bullseye and bookworm's containerd 1.6
  keeps the `ctr -n k8s.io snapshot view` behaviour it relies on.
- Module path renamed; `toolchain go1.24.13` directive dropped so it builds with
  the installed Go rather than fetching its own.
- `gofmt`'d — upstream's tree wasn't, and `check-fmt` covers every `.go` outside
  `./vendor/`.
- `build-preload-disk.sh` loses the sparse-checkout dance; `versions.yaml` held only
  the upstream ref and is gone. Provenance and re-sync instructions are in the
  vendored `README.md`.
- Apache 2.0 `LICENSE` added. Upstream carries per-file headers but no license file,
  and §4(a) wants a copy alongside a redistributed derivative.

Kept as **its own Go module**, so none of its ~30 dependencies reach this repo's
`go.mod` — `git status` confirms the root `go.mod`/`go.sum` are untouched.

## Verified end to end

Real build against `tigera-cc-dev`:

```
Welcome to Debian GNU/Linux 12 (bookworm)!
Setting up containerd
containerd is ready to use
Unpacking is completed
Image has successfully been created at:
  projects/tigera-cc-dev/global/images/cic-vendor-test-131504
```

No 404, no `ctr: command not found`. The image landed in its family and
`describe-from-family` resolved to it; the throwaway was then deleted. Upstream's
own `go test ./...` still passes in the vendored copy.

## Unrelated tidy-up

`strPtr` dropped in favour of `new(x)`, which takes a value as of Go 1.26.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


preload-disks/ci-cache/disk-image-builder/go.mod
hjiawei, line 1: Why a separate go.mod/sum? Does the top-level go.mod/sum work?

Brian-McM: Because we're vendoring it I wanted to ensure the dependencies are the same as what's upstream. I don't want the outdated library to interfere with updating any of our other go binaries.

hjiawei (approved):

